Issue #13, opened 2021-02-22
[High-Fi] Dashboard


gabrielsaruhashi:
**Is your feature request related to a problem? Please describe.**
![image](https://user-images.githubusercontent.com/12942772/108710599-0b46db00-74f3-11eb-8b2c-7104b9f2dcef.png)

**Describe the solution you'd like**

- [ ] Add the empty state in case the user doesn't have any contacts
- [ ] Add the empty state in case the user doesn't have any upcoming calls
- [ ] View up to 3 upcoming calls
- [ ] "See All": Appears in upcoming calls when upcoming calls exceed 3 (4+) —> navigates the user to the "Calls" page

**Additional context**
[Figma ](https://www.figma.com/file/VUyXLtnzCzokVXaEVqaU8s/Connect-Desktop?node-id=20%3A132
)

PR #18, opened 2021-02-25
Lp dashboard

lspdrz:
## Proposed changes

Implements the redesign for Inmate Dashboard

## Related Issues

Closes Issue #13 

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

We still need to do the following:
- Add Spanish translations
- Add Select Options to the dropdown on the CallCancelModal
- Contact the API cancel endpoint upon confirming that modal


src/components/Dashboard/CallDetails.tsx
```diff
@@ -0,0 +1,107 @@
...
+import React from "react";
+import { useTranslation } from "react-i18next";
+import { useAppDispatch } from "src/redux";
+import { Call } from "src/types/Call";
+import { genFullName } from "src/utils/utils";
+import { openCancelCallModal } from "../Modals/modalsSlice";
+
+interface Props {
+  selectedCall: Call | null;
+  onClose: Function;
```
mpekala23, line 22: nit: not a huge thing at all, but I personally am more used to typing functions as `() => void` (or whatever it would be). (Low key did not even know typing it like this was an option.) I slightly prefer `() => void` j because it's more clear a) arguments / return type b) if it's async or not. @gabrielsaruhashi thoughts?

src/components/Dashboard/CallDetails.tsx
```diff
@@ -0,0 +1,107 @@
...
+      new Date(selectedCall.end),
+      new Date(selectedCall.start)
+    );
+
+  return selectedCall ? (
+    <Drawer
+      title="Call Details"
+      placement="right"
+      closable={false}
+      onClose={() => onClose()}
```
mpekala23, line 41: nit: I think you can j do `onClose={onClose}` (also slightly more performant bc doesn't create a function here as prop)

src/components/Dashboard/CallItem.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { push } from "connected-react-router";
+import { differenceInMinutes, format, isToday, isTomorrow } from "date-fns";
+import React from "react";
+import { useTranslation } from "react-i18next";
+import { Call } from "src/types/Call";
+import { genFullName } from "src/utils/utils";
+
+interface Props {
+  call: Call;
+  selectCall: Function;
```
mpekala23, line 11: same comment here re: typing as `(params: type) => returnType`, take or leave

src/components/Dashboard/CallsList.tsx
```diff
@@ -0,0 +1,46 @@
+import { Card, Typography } from "antd";
+import { push } from "connected-react-router";
+import React from "react";
+import { useTranslation } from "react-i18next";
+import { Call } from "src/types/Call";
+import CallItem from "./CallItem";
+
+interface Props {
+  calls: Call[];
+  selectCall: Function;
```
mpekala23, line 10: same here

src/components/Dashboard/ConnectionsList.tsx
```diff
@@ -0,0 +1,92 @@
...
+    },
+    {
+      key: "pending",
+      tab: t("connection.pending"),
+    },
+  ];
+
+  const [activeContactTab, setActiveContactTab] = useState("approved");
+
+  // TODO: Remove this messy function and pass the days past in the api endpoint
```
mpekala23, line 33: is this still needed?

src/components/Modals/CancelCallModal/CancelCallModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { useTranslation } from "react-i18next";
+import { RootState } from "src/redux";
+
+const mapStateToProps = (state: RootState) => ({
+  call: state.modals.cancelCall,
+});
+
+const mapDispatchToProps = { closeModal };
+
+const connector = connect(mapStateToProps, mapDispatchToProps);
```
mpekala23, line 17: nit: I've recently (like maybe a month ago?) started using the `useSelector` and `useDispatch` hooks over this type of const declaration + connector. slightly less code + easier to read imo. In order to more elegantly accept types as well, you can export your own version of these functions called `useTypedSelector` or something that auto-infers types v well

src/components/Modals/Modals.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { connect, ConnectedProps } from "react-redux";
+import modalConfirms from "src/constants/modalConfirms";
+import { RootState } from "src/redux";
+import CancelCallModal from "./CancelCallModal";
+
+const mapStateToProps = (state: RootState) => ({
+  modalId: state.modals.modalId,
+});
+
+const connector = connect(mapStateToProps);
```
mpekala23, line 11: same comment of the `useSelector` hook

src/redux/index.ts
```diff
@@ -24,3 +27,5 @@ export const Store = configureStore({
 });
 
 export type RootState = ReturnType<typeof rootReducer>;
+export type AppDispatch = typeof Store.dispatch;
+export const useAppDispatch = () => useDispatch<AppDispatch>(); // Export a hook that can be reused to resolve types
```
mpekala23, line 31: I think this is what I was talking about re: hooks (there's a similar one for `useSelector`, looks like something related is already included in the inputs?)

mpekala23 (commented):
left some comments! take or leave any of them, looks really good!

src/components/Dashboard/CallDetails.tsx
lspdrz: That makes sense to me!! Especially when considering arguments and return type. I'll go ahead and reimplement across the PR if you agree @gabrielsaruhashi

src/components/Dashboard/ConnectionsList.tsx
lspdrz: Yeah, for now. I made the function for this PR

src/components/Modals/CancelCallModal/CancelCallModal.tsx
lspdrz: YEah, I wanted to discuss this with y'all. Are we trying to impose a standard redux convention where we only implement mapStateToProps instead of hooks or vice-versa? Or can we liberally use either/or when it makes sense?

I like the idea you're getting around above: using the mapState connector for large (parent-like) components only and then use the redux hooks in the smaller components. Thoughts? @gabrielsaruhashi

lspdrz: Also, I was looking in the docs and found that if you define the state in the selector typescript infers for you: https://redux.js.org/recipes/usage-with-typescript#typing-the-useselector-hook

BUT, I don't think this would work if we use useSelector without a selector function (e.g. useSelector(state => state.auth.user). So agreed, I think it's best to just include the new useTypedSelector hook in the root redux file

src/components/Dashboard/CallDetails.tsx
gabrielsaruhashi: yep agreed!

src/components/Modals/CancelCallModal/CancelCallModal.tsx
gabrielsaruhashi: makes sense!

Just gotta make sure we memoize selectors that return new objecst

src/components/Modals/CancelCallModal/CancelCallModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+import Modal from "antd/lib/modal/Modal";
+import { differenceInMinutes, format } from "date-fns";
+import React from "react";
+import { connect, ConnectedProps } from "react-redux";
+import { genFullName } from "src/utils/utils";
+import { closeModal } from "src/components/Modals/modalsSlice";
+import { useTranslation } from "react-i18next";
+import { RootState } from "src/redux";
+
+const mapStateToProps = (state: RootState) => ({
```
gabrielsaruhashi, line 11: @lspdrz , the convetion is to only connect containers to the store.

We should just pass down the id as a props to cancel call modal

src/components/Dashboard/DashboardHeader.tsx
```diff
@@ -0,0 +1,64 @@
...
+            paddingBottom: 0,
+            fontSize: "48px",
+            marginBottom: 0,
+          }}
+        >
+          {format(currTime, "HH:mm")}
+        </Typography.Title>
+        <Typography.Title
+          className="dashboard-header-content"
+          style={{ fontSize: "14px" }}
```
gabrielsaruhashi, line 47: use level={5} instead of hardcoding it

src/components/Dashboard/DashboardHeader.tsx
```diff
@@ -0,0 +1,64 @@
...
+          borderRadius: "8px",
+        }}
+        className="dashboard-header-container banner-background"
+      >
+        <Typography.Title
+          className="dashboard-header-content"
+          style={{
+            paddingTop: 65,
+            paddingBottom: 0,
+            fontSize: "48px",
```
gabrielsaruhashi, line 39: Use level={1} or somt instead of hardcoding

Btw do we need to hardcode padding? 
Cant we use the flexbox or Space

src/components/Modals/modalsSlice.ts
```diff
@@ -0,0 +1,34 @@
+import { createSlice } from "@reduxjs/toolkit";
+import modalConfirms from "src/constants/modalConfirms";
+import { Call } from "src/types/Call";
+
+interface ModalsSliceState {
+  modalId: string;
+  cancelCall: Call | null;
```
gabrielsaruhashi, line 7: I think it'd be cleaner to create some abstract interfaces to avoid having to add a new variable for every new modal

E.g

data: { entity: Call | User | ...; type: string }

and then you create  specific types:

entity: Call, type: call 

to use them on a switch statement

@mpekala23 has some really good examples of this strucutre

gabrielsaruhashi (changes requested):
some suggestions and nits!

src/components/Modals/CancelCallModal/CancelCallModal.tsx
lspdrz: Now doing this with a useSelector

src/components/Dashboard/CallDetails.tsx
```diff
@@ -0,0 +1,107 @@
...
+        </Row>
+        <Row>
+          <Space>
+            <Col>
+              <TeamOutlined style={{ fontSize: "22px" }} />
+            </Col>
+            <Col>
+              <Avatar src={selectedCall.connection.user.profileImgPath} />
+            </Col>
+            <Col style={{ fontSize: 14 }}>
```
gabrielsaruhashi, line 75: let's use tailwind's text-sm

for accessibility and modularity, we should use rem instead of hardcoding values

when we give users the option to adjust their font size, it'll be a matter of just updating the base HTML font size and all rem values will update

src/components/Dashboard/CallDetails.tsx
```diff
@@ -0,0 +1,107 @@
...
+      onClose={onClose}
+      visible={selectedCall ? true : false}
+      getContainer={false}
+      style={{ position: "absolute" }}
+    >
+      <Space direction="vertical" size="large">
+        <Row>
+          <Space>
+            <Col>
+              <CalendarOutlined style={{ fontSize: "22px" }} />
```
gabrielsaruhashi, line 50: same here

src/components/Dashboard/CallsList.tsx
```diff
@@ -0,0 +1,46 @@
...
+      extra={
+        <Typography.Link onClick={() => push("/")}>
+          {t("call.seeAll")}
+        </Typography.Link>
+      }
+      className="rounded-md"
+    >
+      <div style={{ paddingBottom: 16 }}>
+        <Typography.Text>
+          {calls.length === 0
```
gabrielsaruhashi, line 28: Nit: I think it's cleaner to do 

!!calls.length ? 
             `${calls.length} ${t("call.upcomingCalls")}${
                calls.length > 1 ? "s" : ""
              }`   :  t("call.noCalls")

src/components/Dashboard/CallsList.tsx
```diff
@@ -0,0 +1,46 @@
...
+    <Card
+      title={t("call.title")}
+      extra={
+        <Typography.Link onClick={() => push("/")}>
+          {t("call.seeAll")}
+        </Typography.Link>
+      }
+      className="rounded-md"
+    >
+      <div style={{ paddingBottom: 16 }}>
```
gabrielsaruhashi, line 26: we should use tailwind instead of hardcoding padding values

src/components/Dashboard/ConnectionItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+const ConnectionItem: React.FC<Props> = ({ connection }) => {
+  const { t } = useTranslation("dashboard");
+  const endedCalls = useAppSelector(selectEndedCalls);
+
+  // TODO: Remove this messy function and pass the days past in the api endpoint
+  const getDaysPastNum = (connectionId: number) => {
+    if (!endedCalls) return 0;
+    const last = Math.max.apply(
```
gabrielsaruhashi, line 21: wait I think I had a comment on this function, but you can make this one line instead of looping twice over the array.

Check my last PR comments?

src/i18n/es/modals.json
```diff
@@ -0,0 +1,9 @@
+{
+  "cancelCallModal": {
```
gabrielsaruhashi, line 2: Let's just do translations at the end? 

When everything is in

src/components/Dashboard/ConnectionsList.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <Card
+      title={t("connection.title")}
+      tabList={tabList}
+      activeTabKey={activeContactTab}
+      onTabChange={(key) => setActiveContactTab(key)}
+    >
+      <Row justify="space-around">
+        {connections
+          .filter((connection) => connection.status === activeContactTab)
```
gabrielsaruhashi, line 37: nit: for performance, i'd move this to a useState variable and use useEffect to update this array only when connections or tab changes 

Otherwise this will be re-rendered every single time

src/components/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,59 @@
...
+          style={{
+            height: "60px",
+            backgroundColor: "#F0DEFF",
+            opacity: 0.4,
+          }}
+        />
+        <Space
+          direction="vertical"
+          size="large"
+          style={{ padding: 24, paddingTop: 36 }}
```
gabrielsaruhashi, line 30: would use tailwind

src/components/Dashboard/DashboardHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  return (
+    <Space
+      direction="vertical"
+      align="center"
+      style={{
+        backgroundImage: `url(${dailyQuote.background})`,
+      }}
+      className="w-full h-52 rounded-md bg-cover bg-center text-center flex-1 justify-end"
```
gabrielsaruhashi, line 27: v nice!

src/i18n/es/modals.json
```diff
@@ -0,0 +1,9 @@
...
+    "title": "Cancelar llamada?",
+    "okText": "Confirmar",
+    "messageOne": "You are about to cancel your call with {{connectionName}} for:",
+    "messageTwo": "Please select a cancellation reason to send to {{connectionFirstName}}.",
```
gabrielsaruhashi, line 6: oh wow i didnt realize you could use templating language to pass down vars. v cool

src/pages/Dashboard/index.tsx
```diff
@@ -43,205 +19,16 @@ const connector = connect(mapStateToProps, mapDispatchToProps);
...
-                </Row>
-              </Card>
-            </Col>
-          </Row>
-        </Space>
-      </Content>
-    </Layout>
-  );
-}
+  return <Dashboard calls={calls} connections={connections} />;
```
gabrielsaruhashi, line 31: isnt this pulverizing the component way too much?

Wondering if it's best to just have the dashboard component here and importing all components?

@mpekala23 would love to hear ur thoughts

gabrielsaruhashi (changes requested):
Have a couple more comments/nit requests!

src/pages/Dashboard/index.tsx
lspdrz: @gabrielsaruhashi Yep, but later on I would like to pitch implementing Next.js which uses this pages/ directory in a particular way. The pattern implemented here would scale to that

src/components/Dashboard/ConnectionsList.tsx
```diff
@@ -0,0 +1,50 @@
...
+      tab: t("connection.active"),
+    },
+    {
+      key: "pending",
+      tab: t("connection.pending"),
+    },
+  ];
+
+  useEffect(() => {
+    setConnectionItems(
```
gabrielsaruhashi, line 31: 😍

src/components/Modals/CancelCallModal/CancelCallModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+            {t("cancelCallModal.messageTwo", {
+              connectionFirstName: firstName,
+            })}
+          </Typography.Text>
+        </Row>
+        <Row />
+      </Space>
+      <Select
+        placeholder={t("cancelCallModal.dropdownPlaceholder")}
+        style={{ width: "100%" }}
```
gabrielsaruhashi, line 59: nit let's use taiwind?

src/components/Modals/CancelCallModal/CancelCallModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { genFullName } from "src/utils/utils";
+import { closeModal } from "src/components/Modals/modalsSlice";
+import { useTranslation } from "react-i18next";
+import { useAppDispatch, useAppSelector } from "src/redux";
+
+const CancelCallModal: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const call = useAppSelector((state) => state.modals.cancelCall);
+  const { t } = useTranslation("modals");
+  const fullName = call ? genFullName(call.connection.user) : "";
```
gabrielsaruhashi, line 14: Instead of having this check shouldnt we just do

if (!call) return <div /> ?

Then you don't need to check for call on every consecutive line

We might not even want to render the modal if it's going to be fully empty

src/components/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,54 @@
...
+  connections: Connection[];
+}
+const Dashboard: React.FC<Props> = ({ calls, connections }) => {
+  const [selectedCall, setSelectedCall] = useState<null | Call>(null);
+  return (
+    <Layout style={{ minHeight: "100vh" }}>
+      <Content>
+        <div
+          style={{
+            height: "60px",
```
gabrielsaruhashi, line 22: nit: think it'd be good to move to a styles object liked you did on the other file

src/components/Dashboard/CallDetails.tsx
```diff
@@ -0,0 +1,107 @@
...
+      onClose={onClose}
+      visible={selectedCall ? true : false}
+      getContainer={false}
+      style={{ position: "absolute" }}
+    >
+      <Space direction="vertical" size="large">
+        <Row>
+          <Space>
+            <Col>
+              <CalendarOutlined className="text-2xl" />
```
gabrielsaruhashi, line 50: ay!

gabrielsaruhashi (approved):
final nits, but looks INCREDIBLE!

src/components/Modals/CancelCallModal/CancelCallModal.tsx
lspdrz: I agree with this, but I think we should save it for the new modal implementation I made an issue for. Particularly because if we just return an empty react component based on that call value, the parent modal will still render. We'll have to deal with this check on the parent modal when we move the call object there

src/components/Dashboard/Dashboard.tsx
lspdrz: I actually need to abstract this whole div because we're using it in multiple places, so gonna leave like this now

Issue #14, opened 2021-02-22
[High-Fi] Settings page


gabrielsaruhashi:
**Is your feature request related to a problem? Please describe.**
![image](https://user-images.githubusercontent.com/12942772/108729375-350afc80-7509-11eb-8db9-06f132b8aa34.png)

**Describe the solution you'd like**

- [ ] Read resources: I don't have the copy now, but the ability to click on the link and open a modal to read more information would be huge. If the modal could support graphics that could be great but inessential
- [ ] Can change language being used

PR #24, opened 2021-03-04
added settings page

lspdrz:
## Proposed changes

Adds a Settings page to Pathways

## Related Issues

Closes Issue #14 

## Types of changes

What types of changes does your code introduce to Appium?
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

### DIFF
--- a/src/components/Common/Buttons/EditButton/EditButton.tsx
+++ b/src/components/Common/Buttons/EditButton/EditButton.tsx
@@ -1,0 +1,20 @@
+import { SaveOutlined } from "@ant-design/icons";
+import { Button } from "antd";
+import React, { ReactNode } from "react";
+
+interface Props {
+  children?: ReactNode;
+}
+
+const EditButton: React.FC = ({ children }: Props) => (
+  <Button
+    type="primary"
+    icon={<SaveOutlined />}
+    size="large"
+    style={{ borderRadius: 4 }}
+  >
+    {children}
+  </Button>
+);
+
+export default EditButton;
--- a/src/components/Common/Buttons/EditButton/EditButton.tsx
+++ b/src/components/Common/Buttons/EditButton/EditButton.tsx
@@ -11,7 +11,7 @@ const EditButton: React.FC = ({ children }: Props) => (
     type="primary"
     icon={<SaveOutlined />}
     size="large"
-    style={{ borderRadius: 4 }}
+    className="rounded"
   >
     {children}
   </Button>
--- a/src/components/Modals/ResourcesModal/ResourcesModal.tsx
+++ b/src/components/Modals/ResourcesModal/ResourcesModal.tsx
@@ -1,0 +1,23 @@
+import Modal from "antd/lib/modal/Modal";
+import React from "react";
+import { closeModal } from "src/components/Modals/modalsSlice";
+import { useTranslation } from "react-i18next";
+import { useAppDispatch } from "src/redux";
+
+const ResourcesModal: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const { t } = useTranslation("modals");
+
+  return (
+    <Modal
+      title={t("cancelCallModal.title")}
+      visible={true}
+      okText={t("cancelCallModal.okText")}
+      onOk={() => dispatch(closeModal())}
+      onCancel={() => dispatch(closeModal())}
+      className="rounded-sm w-full"
+    ></Modal>
+  );
+};
+
+export default ResourcesModal;
--- a/src/components/Settings/Settings.tsx
+++ b/src/components/Settings/Settings.tsx
@@ -1,26 +1,17 @@
-import React, { useEffect, useState } from "react";
+import React from "react";
 import { Language } from "src/types/Session";
 import PageLayout from "src/components/Common/PageLayout";
 import { Card, Col, Row, Select, Space, Typography } from "antd";
 import { useTranslation } from "react-i18next";
 import { LANGUAGES } from "src/utils/constants";
+import modalConfirms from "src/constants/modalConfirms";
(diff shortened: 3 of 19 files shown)


src/components/Settings/Settings.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const { t, i18n } = useTranslation("settings");
+  const [languageOptions, setLanguageOptions] = useState<
+    JSX.Element[] | undefined
+  >();
+
+  useEffect(() => {
+    const { Option } = Select;
+    const tempLanguageOptions = [];
+    for (const key in LANGUAGES) {
+      tempLanguageOptions.push(
```
gabrielsaruhashi, line 18: oh hmm bc it's just JSX mapping, I don't think we need to assign it to a var

I'd just use map() directly on the return line and make it one line

React is really good at re-rendering strategies for components (that's why it requires us to add keys to arrays)

src/components/Settings/Settings.tsx
```diff
@@ -0,0 +1,78 @@
...
+              <div className="pt-4">
+                <Typography.Text>{`${t("language.title")}:`}</Typography.Text>
+                {/* TODO: Right now, this only changes the language on the frontend via i18n. We need to also update the database
+                                whenever we have an API endpoint for it */}
+                <Select
+                  className="pl-4"
+                  defaultValue={i18n.language}
+                  onChange={(value) => i18n.changeLanguage(value)}
+                >
+                  {languageOptions}
```
gabrielsaruhashi, line 48: Think you can just map directly here and make it one line with LANGUAGES.map()

src/components/Settings/Settings.tsx
```diff
@@ -0,0 +1,78 @@
...
+                </Select>
+              </div>
+            </Card>
+          </Col>
+          <Col span={12}>
+            <Card title={t("resources.title")}>
+              <Row className="pb-2">
+                <Typography.Link href="/" target="_blank">
```
gabrielsaruhashi, line 56: I think I copied Liz's notes, but it might be good here to already scaffold the modal that's gonna open with the FAQ answer. 

Check the issue for details on the interaction?

gabrielsaruhashi (changes requested):
🚀 looks great ! just some nits
alt

src/components/Settings/Settings.tsx
```diff
@@ -0,0 +1,83 @@
...
+                <Typography.Text>{`${t("language.title")}:`}</Typography.Text>
+                {/* TODO: Right now, this only changes the language on the frontend via i18n. We need to also update the database
+                                whenever we have an API endpoint for it */}
+                <Select
+                  className="pl-4"
+                  defaultValue={i18n.language}
+                  onChange={(value) => i18n.changeLanguage(value)}
+                >
+                  {Object.keys(LANGUAGES).map((key) => (
+                    <Option value={key}>{LANGUAGES[key as Language]}</Option>
```
gabrielsaruhashi, line 40: I think you're missing the "key" props here 
React is gonna complain

src/components/Settings/Settings.tsx
```diff
@@ -0,0 +1,83 @@
...
+                <Typography.Text>{t("language.messageOne")}</Typography.Text>
+              </p>
+              <p>
+                <Typography.Text type="secondary">
+                  {t("language.messageTwo")}
+                </Typography.Text>
+              </p>
+              <div className="pt-4">
+                <Typography.Text>{`${t("language.title")}:`}</Typography.Text>
+                {/* TODO: Right now, this only changes the language on the frontend via i18n. We need to also update the database
```
gabrielsaruhashi, line 32: oh why would we need to update it on the database as well?

src/components/Modals/ResourcesModal/ResourcesModal.tsx
```diff
@@ -0,0 +1,23 @@
+import Modal from "antd/lib/modal/Modal";
+import React from "react";
+import { closeModal } from "src/components/Modals/modalsSlice";
+import { useTranslation } from "react-i18next";
+import { useAppDispatch } from "src/redux";
+
+const ResourcesModal: React.FC = () => {
+  const dispatch = useAppDispatch();
```
gabrielsaruhashi, line 8: I think I made this comment on the other modal, but wouldn't it be cleaner to pass down
closeModal from the Modal container instead of having to connect all these child components to the store?
It'd be nice to keep all child components dummy

src/components/Common/Buttons/EditButton/EditButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+interface Props {
+  children?: ReactNode;
+}
+
+const EditButton: React.FC = ({ children }: Props) => (
+  <Button
+    type="primary"
+    icon={<SaveOutlined />}
+    size="large"
+    style={{ borderRadius: 4 }}
```
gabrielsaruhashi, line 14: tailwind?

gabrielsaruhashi (approved):
some more questions/suggestions, but will stamp it!

src/components/Settings/Settings.tsx
lspdrz: Just so users don't have to select a language every time they log in. Alternatively, we could save it as a cookie

src/components/Modals/ResourcesModal/ResourcesModal.tsx
lspdrz: Let's talk about this! I think both patterns would be about the same

PR #29, opened 2021-03-25
Lp calls

lspdrz:
FIXES THE CALLS API INTEGRATION

### DIFF
--- a/src/api/Common.ts
+++ b/src/api/Common.ts
@@ -56,6 +56,10 @@ export async function fetchAuthenticated(
 
   const body = await response.json();
 
+  if (!body.data) {
+    throw body;
+  }
+
   return body;
 }
 
--- a/src/api/Common.ts
+++ b/src/api/Common.ts
@@ -50,15 +50,11 @@ export async function fetchAuthenticated(
     timeout
   );
 
-  if (response.status !== 200) {
+  if (response.status !== 200 && response.status !== 201) {
     throw response;
   }
 
   const body = await response.json();
-
-  if (!body.data) {
-    throw body;
-  }
 
   return body;
 }
--- a/src/api/Common.ts
+++ b/src/api/Common.ts
@@ -69,7 +69,7 @@ export async function fetchFacilities(
     },
   };
   const response = await fetchTimeout(
-    `${API_URL}facilities`,
+    `${API_URL}public/facilities`,
     requestOptions,
     timeout
   );
--- a/src/components/Calls/Calls.tsx
+++ b/src/components/Calls/Calls.tsx
@@ -1,6 +1,6 @@
 import { Avatar, Button, Col, Row, Space, Table, Typography } from "antd";
 import React, { useEffect, useState } from "react";
-import { Call } from "src/types/Call";
+import { BaseCall, Call } from "src/types/Call";
 import { User } from "src/types/User";
 import PageLayout from "src/components/Common/PageLayout";
 import { useTranslation } from "react-i18next";
@@ -10,7 +10,7 @@ import { useAppDispatch } from "src/redux";
 import { openModal } from "src/redux/modules/modalsSlice";
 
 interface Props {
-  calls: Call[];
+  calls: BaseCall[];
   user: User;
 }
 
@@ -41,6 +41,8 @@ const tableColumns = [
 ];
 
 const Calls: React.FC<Props> = ({ calls, user }) => {
+  console.log("hey from calls");
+  console.log(calls);
   const dispatch = useAppDispatch();
   const [tableData, setTableData] = useState<TableData[] | undefined>(
     undefined
@@ -51,11 +53,11 @@ const Calls: React.FC<Props> = ({ calls, user }) => {
     setTableData(
       calls.map((call) => {
         return {
-          date: format(new Date(call.start), "eeee, LLLL d"),
-          time: `${format(new Date(call.start), "h:mm bbb")} - ${format(
-            new Date(call.start),
+          date: format(new Date(call.scheduledStart), "eeee, LLLL d"),
+          time: `${format(
+            new Date(call.scheduledStart),
             "h:mm bbb"
-          )}`,
+          )} - ${format(new Date(call.scheduledStart), "h:mm bbb")}`,
           participants: (
             <Space>
               <Avatar src={user.profileImagePath} size="small" />
--- a/src/components/Calls/Calls.tsx
+++ b/src/components/Calls/Calls.tsx
@@ -41,8 +41,6 @@ const tableColumns = [
 ];
 
 const Calls: React.FC<Props> = ({ calls, user }) => {
-  console.log("hey from calls");
-  console.log(calls);
   const dispatch = useAppDispatch();
   const [tableData, setTableData] = useState<TableData[] | undefined>(
     undefined
--- a/src/components/Calls/Calls.tsx
+++ b/src/components/Calls/Calls.tsx
@@ -1,6 +1,6 @@
 import { Avatar, Button, Col, Row, Space, Table, Typography } from "antd";
 import React, { useEffect, useState } from "react";
-import { BaseCall, Call } from "src/types/Call";
+import { BaseCall } from "src/types/Call";
 import { User } from "src/types/User";
 import PageLayout from "src/components/Common/PageLayout";
 import { useTranslation } from "react-i18next";
--- a/src/components/Dashboard/CallItem.tsx
+++ b/src/components/Dashboard/CallItem.tsx
@@ -21,11 +21,11 @@ const CallItem: React.FC<Props> = ({
   const { t } = useTranslation("dashboard");
 
   const duration = differenceInMinutes(
-    new Date(call.end),
-    new Date(call.start)
+    new Date(call.scheduledEnd),
+    new Date(call.scheduledStart)
   );
 
-  const started = new Date(call.start) < new Date();
+  const started = new Date(call.scheduledStart) < new Date();
 
   const getDateLabel = (date: Date) => {
     if (isToday(date)) return "Today";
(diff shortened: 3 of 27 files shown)


src/components/Calls/Calls.tsx
```diff
@@ -1,6 +1,6 @@
 import { Avatar, Button, Col, Row, Space, Table, Typography } from "antd";
 import React, { useEffect, useState } from "react";
-import { Call } from "src/types/Call";
+import { BaseCall, Call } from "src/types/Call";
```
gabrielsaruhashi, line 3: delete unusued Call type import?
It's probably throwing a warning

src/redux/modules/connection.ts
```diff
@@ -11,6 +11,8 @@ export const fetchConnections = createAsyncThunk(
   "connections/fetchAll",
   async () => {
     const body = await fetchAuthenticated(`connections`);
+    console.log("hey");
```
gabrielsaruhashi, line 14: delete?

src/redux/modules/call.ts
```diff
@@ -13,10 +13,13 @@ import { showToast } from "src/utils";
 
 export const fetchCalls = createAsyncThunk("calls/fetchAll", async () => {
   const body = await fetchAuthenticated(`calls`);
-  if (body.status !== 200) {
+  if (!body.data) {
     throw body;
   }
 
+  console.log("hey calls");
```
gabrielsaruhashi, line 20: delete

src/redux/modules/call.ts
```diff
@@ -13,10 +13,13 @@ import { showToast } from "src/utils";
 
 export const fetchCalls = createAsyncThunk("calls/fetchAll", async () => {
   const body = await fetchAuthenticated(`calls`);
-  if (body.status !== 200) {
+  if (!body.data) {
```
gabrielsaruhashi, line 16: just move this to the fetchauthenticated call
I sent mark a screenshot on connect tech

src/redux/selectors.ts
```diff
@@ -26,7 +26,7 @@ const getCallEntities = (
   // const kiosk = selectKioskById(state, call.kioskId);
   // if (!kiosk) return;
 
-  return { ...call, connection };
+  return { ...call };
```
gabrielsaruhashi, line 29: just 
`return call `

src/types/Connection.ts
```diff
@@ -8,4 +8,9 @@ export interface Connection {
   statusDetails: string;
   relationship: string;
   user: Contact;
+  lastCall: {
```
gabrielsaruhashi, line 11: what is this for?
I think I commented on the backend PRs (if you replied, could you bump the thread? ), but why do this calculation server side? 
We should let clients do this expensive computation

src/types/Call.ts
```diff
@@ -1,21 +1,23 @@
...
 export interface BaseCall {
   id: number;
-  start: number;
-  end: number;
+  scheduledStart: Date;
+  scheduledEnd: Date;
   approved: boolean;
   status: "ended" | "scheduled" | "terminated" | "rescheduled" | "live";
   connectionId: number;
+  userParticipants: User[];
```
gabrielsaruhashi, line 15: isnt the backend returning a userId array?
This should just be the inmateIds

src/types/Call.ts
```diff
@@ -1,21 +1,23 @@
...
   kioskId: number;
 }
 
 export interface Call extends BaseCall {
-  connection: Connection;
+  // connection: Connection;
```
gabrielsaruhashi, line 20: and then here we have the users.
The point of this setup is to ensure data normalization. We already have this data from the connection redux model

src/redux/selectors.ts
```diff
@@ -26,7 +26,7 @@ const getCallEntities = (
   // const kiosk = selectKioskById(state, call.kioskId);
   // if (!kiosk) return;
```
gabrielsaruhashi, line 27: you should be selecting the user from the connection array here

src/pages/Login/index.tsx
```diff
@@ -57,7 +57,7 @@ function LoginContainer({ session }: PropsFromRedux): ReactElement {
   useEffect(() => {
     const getFacilityOptions = async () => {
       const response = await fetchFacilities();
-      const facilities = response.data;
+      const facilities = response.data.results;
       const facilityOptions = facilities.map((facility) => (
         <Option value={facility.id}>{facility.name}</Option>
```
gabrielsaruhashi, line 62: wait why keep this here?
It's odd to have a component be the dependency array - this is not a use case for useeffect imo

I'd just load the facilities data upon mounting on use effect
and then u render options.
We shouldn't be storing JSX elements in a useState variable

src/components/Dashboard/CallItem.tsx
```diff
@@ -38,15 +38,16 @@ const CallItem: React.FC<Props> = ({
...
+            {getDateLabel(new Date(call.scheduledStart))}
           </Typography.Title>
           <Typography.Text>
-            {format(new Date(call.start), "h:mm aaa")} • {`${duration} minutes`}
+            {format(new Date(call.scheduledStart), "h:mm aaa")} •{" "}
+            {`${duration} minutes`}
           </Typography.Text>
           <Space style={{ paddingTop: 18 }}>
-            <Avatar src={call.connection.user.profileImagePath} />
+            <Avatar src={call.userParticipants[0].profileImagePath} />
```
gabrielsaruhashi, line 48: I think this a bit too unsafe...
I'd just map over the userparticipants array and render the Avatars

Also there might be more than two user participants registered for a call

src/api/Common.ts
```diff
@@ -69,7 +69,7 @@ export async function fetchFacilities(
     },
   };
   const response = await fetchTimeout(
-    `${API_URL}facilities`,
+    `${API_URL}public/facilities`,
```
gabrielsaruhashi, line 72: maybe use that new URL() method?

gabrielsaruhashi (changes requested):
some suggestions!

src/types/Connection.ts
lspdrz: Nah, we should always do expensive computation stuff on the server side. it's so much faster than on the browser and won't affect experience.

Good catch though because this backend change was on the first PR I did last week, I need to include that code in a new backend PR

src/types/Call.ts
lspdrz: Nah, it's returning the User objects which we need on the frontend

src/types/Call.ts
lspdrz: I don't understand here

src/components/Dashboard/CallItem.tsx
lspdrz: Do we want to render all of the participants or just one? The designs only include one participant on the dashboard, but maybe that's because of our earlier assumption there would only ever be one connection

src/api/Common.ts
lspdrz: I tried this for a while and had a problem, I don't remember what they were. I can try again if necessary

src/pages/Login/index.tsx
```diff
@@ -87,6 +84,10 @@ function LoginContainer({ session }: PropsFromRedux): ReactElement {
     showToast("login_error", "Invalid ID or Pin Code", "error");
   };
 
+  const facilityOptions = facilities.map((facility: FacilityRO) => (
+    <Option value={facility.id}>{facility.name}</Option>
```
gabrielsaruhashi, line 88: you're missing a key here
react is gonna throw a warning/error bc they use these keys to optimize the rerendering strategy

src/api/Common.ts
```diff
@@ -56,6 +56,10 @@ export async function fetchAuthenticated(
 
   const body = await response.json();
 
+  if (!body.data) {
```
gabrielsaruhashi, line 59: wait why add this check?
I think response.status is enough

gabrielsaruhashi, line 59: we should be checking for 200 and 201 btw

POST's return a 201 status code

src/redux/selectors.ts
gabrielsaruhashi: the goal of this function is to retrieve the relationships of the call object.

that's why I was fetching the connection object. We use redux-toolkit entity adapter to normalize the data instead of duplicating it across multiple objects

e.g how are you currently getting the names of the participants? getCallEntities should iterate over the list of ids of outside users that will be on the call  to grab their information

gabrielsaruhashi (changes requested):
more feedback!

src/components/Dashboard/ConnectionsList.tsx
```diff
@@ -2,15 +2,15 @@ import { Card, Row } from "antd";
...
 interface Props {
   calls: Call[];
-  connections: Connection[];
+  contacts: Contact[];
 }
 
-const ConnectionsList: React.FC<Props> = ({ calls, connections }: Props) => {
+const ConnectionsList: React.FC<Props> = ({ calls, contacts }: Props) => {
   const { t } = useTranslation("dashboard");
   const [connectionItems, setConnectionItems] = useState<JSX.Element[] | null>(
```
gabrielsaruhashi, line 15: ok same comment here but we shouldnt be storing JSX in use state

this is what happens underneath the hood: https://reactjs.org/docs/lists-and-keys.html

src/redux/modules/contactsSlice.ts
```diff
@@ -0,0 +1,41 @@
...
+  return contacts;
+});
+
+export const contactsSlice = createSlice({
+  name: "contacts",
+  initialState: contactsAdapter.getInitialState(),
+  reducers: {},
+  extraReducers: (builder) => {
+    builder.addCase(fetchContacts.fulfilled, (state, action) => {
+      console.log("hi here in contact adapter");
```
gabrielsaruhashi, line 26: delete?

src/redux/selectors.ts
```diff
@@ -11,22 +12,22 @@ export const {
...
+  selectAll: selectAllContacts,
+  selectById: selectContactById,
+} = contactsAdapter.getSelectors<RootState>((state) => state.contacts);
 
 const getCallEntities = (
   state: RootState,
   call: BaseCall
 ): Call | undefined => {
-  const connection = selectConnectionById(state, call.connectionId);
+  const connection = selectContactById(state, call.connectionId);
```
gabrielsaruhashi, line 23: nit: change to contact?

src/redux/selectors.ts
```diff
@@ -11,22 +12,22 @@ export const {
...
   if (!connection) return;
 
   // TODO add kiosk info soon
   // const kiosk = selectKioskById(state, call.kioskId);
   // if (!kiosk) return;
 
-  return { ...call, connection };
+  return call;
```
gabrielsaruhashi, line 30: u need to return the contact here right?
rn u'r fetching the contact but not returning it with the call inffo

src/types/Connection.ts
```diff
@@ -1,11 +1,13 @@
...
-
 export interface Connection {
   id: number;
   userId: number;
   inmateId: number;
   status: string;
   statusDetails: string;
   relationship: string;
-  user: Contact;
+  lastCall: {
```
gabrielsaruhashi, line 8: why not use the Call object here?

src/types/User.ts
```diff
@@ -15,4 +17,12 @@ export interface User extends BasePersona {
 
 export interface Contact extends BasePersona {
   email: string;
+  status: string;
+  statusDetails: string;
+  relationship: string;
+  lastCall: {
```
gabrielsaruhashi, line 23: why not use the call object here?

gabrielsaruhashi (changes requested):
last round of feedback!

PR #43, opened 2021-04-21
integrated malan

lspdrz:


mpekala23 (approved):
looks good to me! sorry copied the code to mobile but forgot to stamp.

you could probably get rid of the cookie comment if you wanted (in `api/User.ts`)

PR #56, opened 2021-07-26
In-Person Visits

lspdrz:
New IPV Feature


src/pages/Dashboard/index.tsx
```diff
@@ -6,31 +6,34 @@ import { push } from "connected-react-router";
...
     return () => clearInterval(interval);
   }, [dispatch]);
 
   return (
     <Dashboard
-      calls={calls}
+      meetings={meetings}
       contacts={contacts}
-      joinCall={(call: Call) => {
+      joinCall={(meeting: Meeting) => {
```
gabrielsaruhashi, line 34: does it make sense for joinCall to take a Meeting as the argument?
Shouldnt it just be Call?

src/pages/Dashboard/index.tsx
```diff
@@ -6,31 +6,34 @@ import { push } from "connected-react-router";
...
-  const calls = useUpcomingCalls();
+  const meetings = useUpcomingVisits();
 
   useEffect(() => {
     dispatch(fetchCalls());
+    dispatch(fetchInPersonVisits());
 
     const interval = setInterval(() => {
       dispatch(fetchCalls());
+      dispatch(fetchInPersonVisits());
```
gabrielsaruhashi, line 25: oh I dont think you need to included this here.
The only reason I have an interval here is in case the video handler information changes for the call. This wont happen for in person visits

src/types/InPersonVisit.ts
```diff
@@ -0,0 +1,8 @@
+import { Contact } from "./User";
+import { BaseVisit } from "./Visit";
+
+export interface BaseInPersonVisit extends BaseVisit {}
```
gabrielsaruhashi, line 4: is this being used?

src/types/Meeting.ts
```diff
@@ -0,0 +1,79 @@
...
+  | InPersonContactSlot
+  | InPersonNoContactSlot;
+
+export interface BaseMeeting {
+  id: string;
+  facilityId: string;
+  status: MeetingStatus;
+  statusDetails: string | null;
+  kioskName: string;
+  scheduledStart: string;
```
gabrielsaruhashi, line 48: which one is being use? Meeting or Call.ts?

src/types/Meeting.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export interface InPersonNoContactVisit extends BaseMeeting {
+  type: "inPersonNoContact";
+}
+
+export type InPersonVisit = InPersonContactVisit | InPersonNoContactVisit;
+
+export type Meeting = Call | InPersonVisit;
+
+type CustomFilter = string & { dummy?: never };
```
gabrielsaruhashi, line 73: I dont think you need some of these types - cleanup?

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+interface Props {
+  meetings: Meeting[];
+  user: User;
+}
+
+const Meetings: React.FC<Props> = ({ meetings, user }) => {
+  const dispatch = useAppDispatch();
+  const { t } = useTranslation("calls");
+
+  const [meetingCategory, setMeetingCategory] = useState<
```
gabrielsaruhashi, line 37: would rename this meetingTypeFilter to make it more clear what it does

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+    "call" | "inPersonContact" | "inPersonNoContact" | ""
+  >("");
```
gabrielsaruhashi, line 39: would make this nullable, makes more sense with the filter implementation

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+  const [selectedContacts, setSelectedContacts] = useState<string[]>([]);
```
gabrielsaruhashi, line 40: selectedContactsFilter

gabrielsaruhashi: would make this nullable in case the filter is not being used

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+        return <PeopleGray />;
+      default:
+        return <VideoCameraGray />;
+    }
+  };
+
+  const filterVisits = () => {
+    return meetings.filter((meeting) => {
+      let isOk = true;
+      if (meetingCategory.length !== 0) {
```
gabrielsaruhashi, line 59: check comments above! but if we make it nullable we wont have to make these awk length checks

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+        isOk = meeting.type === meetingCategory;
+      }
+      if (selectedContacts.length !== 0) {
+        const name = `${meeting.userParticipants[0].firstName} ${meeting.userParticipants[0].lastName}`;
```
gabrielsaruhashi, line 63: Why hardcode the 0 index ? 
We should map over all the participants. We're gonna add extra use rparticipants on the next sprint

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+  meetings: Meeting[];
+  user: User;
+}
+
+const Meetings: React.FC<Props> = ({ meetings, user }) => {
+  const dispatch = useAppDispatch();
+  const { t } = useTranslation("calls");
+
+  const [meetingCategory, setMeetingCategory] = useState<
+    "call" | "inPersonContact" | "inPersonNoContact" | ""
```
gabrielsaruhashi, line 38: would create a type on Meeting.ts callled MeetingType so we don't need to type this out in multiple placecs

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+      default:
+        return <VideoCameraGray />;
+    }
+  };
+
+  const filterVisits = () => {
+    return meetings.filter((meeting) => {
+      let isOk = true;
+      if (meetingCategory.length !== 0) {
+        isOk = meeting.type === meetingCategory;
```
gabrielsaruhashi, line 60: why not just return false here if the meeting does not pass this check?
Right now this makes these filter OR operators, but we probably want AND checks

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+      }
+      if (selectedContacts.length !== 0) {
+        const name = `${meeting.userParticipants[0].firstName} ${meeting.userParticipants[0].lastName}`;
+        isOk = selectedContacts.includes(name);
```
gabrielsaruhashi, line 64: same comment here re: returning false if it doesn't pass the active contact filter

We would then get rid of this mutable isOk variable

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+    },
+    {
+      title: "",
+      key: "cancel",
+      render: (meeting: Meeting) => (
+        <Button
+          type="link"
+          onClick={() =>
+            dispatch(
+              openModal({ activeType: "CANCEL_CALL_MODAL", entity: meeting })
```
gabrielsaruhashi, line 149: cancel meeting modal?

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+    },
+  ];
+
+  const eventTypesMenu = (
+    <Menu>
+      <Menu.Item>
+        <Row align="middle">
+          <Button
+            type="text"
+            style={{ paddingLeft: "0px" }}
```
gabrielsaruhashi, line 165: tailwind

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+            onClick={() =>
```
gabrielsaruhashi, line 166: move this to a Chip component? 
Seems like it's repeated logic

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+                </Space>
+                <Space direction="vertical">
+                  <Typography.Text type="secondary">
+                    {"Daily visit limit"}
+                  </Typography.Text>
+                  <Space>
+                    <Typography.Text strong className="text-base">
+                      {user.inPersonVisitQuota}
+                    </Typography.Text>
+                    {/* <Tooltip title="PLACEHOLDER FOR liz">
```
gabrielsaruhashi, line 276: we can delete! mb think i lef this in tthere

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+                    <Button icon={InfoOutlined} shape="circle" />
+                  </Tooltip> */}
+                  </Space>
+                </Space>
+              </Space>
+            </Tooltip>
+          </Col>
+        </Row>
+        <Row align="middle">
+          <Typography.Text type="secondary" style={{ paddingBottom: "3px" }}>
```
gabrielsaruhashi, line 286: tailwind

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+            {"View: "}
+          </Typography.Text>
+          <Dropdown overlay={eventTypesMenu}>
+            <Button type="link" className="ant-dropdown-link">
```
gabrielsaruhashi, line 290: do we need to hardcode the antd classname?

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+            In-Person Visit, No Contact
+          </Button>
+          {meetingCategory === "inPersonNoContact" ? <CheckOutlined /> : null}
+        </Row>
+      </Menu.Item>
+    </Menu>
+  );
+
+  const contactTypesMenu = () => {
+    let contacts = meetings.map(
```
gabrielsaruhashi, line 213: there should be a util function in Call.ts that creates fulll names for all participants
if not we should probably move it there

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,316 @@
...
+      (meeting) =>
+        `${meeting.userParticipants[0].firstName} ${meeting.userParticipants[0].lastName}`
+    );
+    contacts = contacts.filter(
```
gabrielsaruhashi, line 217: im a bit confused by this logic?
self is the contacts array itself, right?
what is this check doing? 
seeems like it will return true for everything

src/components/Dashboard/MeetingsList.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+const MeetingsList: React.FC<Props> = ({
+  meetings,
+  selectMeeting,
+  joinCall,
+  seeAllCalls,
+}: Props) => {
+  const { t } = useTranslation("dashboard");
+  const [filteredMeetings, setFilteredMeetings] = useState<Meeting[]>([]);
+  const [activeMeetingTab, setActiveMeetingTab] = useState("allTypes");
```
gabrielsaruhashi, line 23: would type it to ensure it's safe
using the keys you created for tablist

gabrielsaruhashi: can be a type we declare just for this page
no need to export it from somewhere else imoo

src/components/Dashboard/MeetingsList.tsx
```diff
@@ -0,0 +1,129 @@
...
+      const values = date.split(",");
+      const scheduledStart = values[0];
+      const weekDay = weekDays[parseInt(values[1])];
+
+      const dayOfMonth = new Date(scheduledStart).getDate();
+      const meetingsAtDate = filteredMeetings.filter(
+        (meeting) =>
+          new Date(meeting.scheduledStart).getDay() ===
+          new Date(scheduledStart).getDay()
+      );
```
gabrielsaruhashi, line 85: nit: add whitespace between return statement and function?

src/components/Dashboard/MeetingsList.tsx
```diff
@@ -0,0 +1,129 @@
...
+              <Typography.Title level={5} type="secondary">
+                {weekDay.toUpperCase()}
+              </Typography.Title>
+              <Typography.Text strong>{dayOfMonth}</Typography.Text>
+            </Space>
+          </Col>
+          <Col flex="auto">
+            {meetingsAtDate.slice(0, 5).map((meeting) => (
+              <MeetingItem
+                meeting={meeting as Meeting}
```
gabrielsaruhashi, line 99: why do we need to force the typing here?

src/components/Dashboard/MeetingsList.tsx
```diff
@@ -0,0 +1,129 @@
...
+          )
+        );
+        break;
+    }
+  }, [activeMeetingTab, meetings]);
+
+  const getMeetingItems = () => {
+    const meetingsByDate = filteredMeetings.filter(
+      (meeting, index, self) =>
+        index ===
```
gabrielsaruhashi, line 64: a bit confused by this logic, not sure I understand what this self.findINdex is achieving -- there are two nested for loops? This would be n^2 right? 

but couldn't we just `groupBy` ahead of time and filter accordingly?

I have a groupby implementation in utils in the connect-client repo. I think it should be in this repo as well, but if not, you can just find it there

src/components/Dashboard/MeetingsList.tsx
```diff
@@ -0,0 +1,129 @@
...
+          (m) =>
+            new Date(m.scheduledStart).getDay() ===
+            new Date(meeting.scheduledStart).getDay()
+        )
+    );
+    const dates = meetingsByDate.map((meeting) => {
+      const date = new Date(meeting.scheduledStart);
+      return `${new Date(meeting.scheduledStart)}, ${date.getDay()}`;
+    });
+    return dates.map((date) => {
```
gabrielsaruhashi, line 75: coudlnt we just do this alll in one map instead of two consecutive ones?

gabrielsaruhashi (changes requested):
have some questions and suggestions!

src/components/Modals/Modals.tsx
```diff
@@ -27,6 +28,9 @@ const Modals: React.FC = () => {
           cancelCall={(id: string, reason: string) =>
             dispatch(cancelCall({ id, reason }))
           }
+          cancelInPersonVisit={(id: string, reason: string) =>
```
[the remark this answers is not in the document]
lspdrz: It is the same modal that we use to cancel not both a call and in-person visit, but either a call or in-person-visit

src/components/Dashboard/MeetingItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+        return <PeopleBurgundy style={{ marginBottom: "10px" }} />;
+      default:
+        return <VideoCamera style={{ marginBottom: "10px" }} />;
+    }
+  };
+
+  const getLeftBorderColor = () => {
+    switch (meeting.type) {
+      case "call":
+        return "#569FE1";
```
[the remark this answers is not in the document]
lspdrz: Do we have an example of where we're using tailwind colors?

src/components/Dashboard/MeetingItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+              type="primary"
+              className="rounded-sm"
+              disabled={
+                (meeting.type === "call" && !meeting.videoHandler) || !canJoin
+              }
+              onClick={() => {
+                joinCall(meeting);
+              }}
+            >
+              {meeting.type === "call" && meeting.videoHandler && canJoin
```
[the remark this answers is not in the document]
lspdrz: I don't think it's a case statement, there's nested conditionals here

src/types/Meeting.ts
lspdrz: Meeting

src/types/Meeting.ts
lspdrz: i'd rather keep for consistency across codebases if needed?

src/components/Meetings/Meetings.tsx
lspdrz: We want to check both filters

src/components/Meetings/Meetings.tsx
lspdrz: chip component?

src/components/Meetings/Meetings.tsx
lspdrz: yeah that's what the docs did

src/components/Meetings/Meetings.tsx
lspdrz: It's deleting all the duplicate values from the array

src/components/Dashboard/MeetingsList.tsx
lspdrz: no longer needed

src/components/Dashboard/MeetingsList.tsx
lspdrz: It's getting all of dates for the meetings, without any duplicates

src/components/Dashboard/MeetingsList.tsx
lspdrz: I can't type, it messes with antd

src/types/Call.ts
```diff
@@ -28,22 +27,13 @@ export type CallStatus = InCallStatus | GeneralCallStatus;
...
-  statusDetails: string;
-  userIds: string[];
-  kioskId: string;
-  kioskName: string;
   videoHandler?: CallHandler;
 }
 
 export interface Call extends BaseCall {
-  userParticipants: Contact[];
+  // userParticipants: Contact[];
```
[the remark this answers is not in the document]
lspdrz: Can you elaborate here?

src/components/Dashboard/MeetingItem.tsx
```diff
@@ -0,0 +1,123 @@
...
+    }
+  };
+
+  return (
+    <Card
+      key={meeting.id}
+      style={{
+        borderLeftColor: getLeftBorderColor(),
+        borderLeftWidth: 3,
+        marginBottom: "16px",
```
gabrielsaruhashi, line 68: tailwind

gabrielsaruhashi: instead of harcoding a margin bottom on the meeting item level, I'd just add spacing on the parent div that wraps this list of items

src/components/Dashboard/MeetingItem.tsx
```diff
@@ -0,0 +1,123 @@
...
+          </Typography.Text>
+          <Space className="mt-4">
+            <ContactAvatarGroup contacts={meeting.userParticipants} />
+            <Typography.Text type="secondary">
+              {getParticipantsFullNames(meeting)}
+            </Typography.Text>
+          </Space>
+        </Space>
+        <Space>
+          {isInitializing ? (
```
gabrielsaruhashi, line 91: let's move this to a separate function?
it's starting to get pretty complicated

Instead of having a disabled "join" button for in person visits, I'd just make it so that "view" details is always the active button for IPV.

src/components/Dashboard/MeetingsList.tsx
gabrielsaruhashi: bumping request

src/components/Dashboard/MeetingsList.tsx
```diff
@@ -0,0 +1,130 @@
...
+      const scheduledStart = values[0];
+      const weekDay = weekDays[parseInt(values[1])];
+
+      const dayOfMonth = new Date(scheduledStart).getDate();
+      const meetingsAtDate = filteredMeetings.filter(
+        (meeting) =>
+          new Date(meeting.scheduledStart).getDay() ===
+          new Date(scheduledStart).getDay()
+      );
+      return (
```
gabrielsaruhashi, line 87: could we move this into its own component?

src/components/Dashboard/MeetingsList.tsx
```diff
@@ -0,0 +1,130 @@
...
+              meeting.meetingType === "in_person_contact" ||
+              meeting.meetingType === "in_person_no_contact"
+          )
+        );
+        break;
+    }
+  }, [activeMeetingTab, meetings]);
+
+  const getMeetingItems = () => {
+    const meetingsByDate = filteredMeetings.filter(
```
gabrielsaruhashi, line 64: could we type this variable to make it easier to read?

src/components/Meetings/Meetings.tsx
gabrielsaruhashi: bump

src/components/Meetings/Meetings.tsx
gabrielsaruhashi: I'd move it to its own component and call it Chip or something. It's being repeated 3 times rn

src/components/Meetings/Meetings.tsx
gabrielsaruhashi: wait why would there be a duplicate in the contact array?

src/types/Meeting.ts
gabrielsaruhashi: what is the use case of this dummy: never?

src/components/Modals/Modals.tsx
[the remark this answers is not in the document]
gabrielsaruhashi: why not just have cancelMeeting here? 
cancelInPersonVisit and CancelMeetingModal both use the same endpoint don't they? 
And they call the same meeting ID

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,327 @@
...
+  };
+
+  return (
+    <PageLayout>
+      <Space className="p-8 w-full" direction="vertical">
+        <Row className="pb-2">
+          <Col>
+            <Tooltip title={t("tooltip")}>
+              <Space>
+                <Space direction="vertical" style={{ paddingRight: "36px" }}>
```
gabrielsaruhashi, line 272: tailwind

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,327 @@
...
+      }
+      if (selectedContactsFilter) {
+        const name = `${meeting.userParticipants[0].firstName} ${meeting.userParticipants[0].lastName}`;
+        isOk = selectedContactsFilter.includes(name);
+      }
+      return isOk;
+    });
+  };
+
+  const tableColumns = [
```
gabrielsaruhashi, line 73: can we move the table to its own component? 
This file is a bit lengthy rn which makes it hard to read

src/components/Meetings/Meetings.tsx
gabrielsaruhashi: bump

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,327 @@
...
+  ] = useState<MeetingType | null>(null);
+  const [selectedContactsFilter, setSelectedContactsFilter] = useState<
+    string[] | null
+  >(null);
+  const [selectedMeeting, setSelectedMeeting] = useState<null | Meeting>(null);
+
+  const getIcon = (visitType: MeetingType) => {
+    switch (visitType) {
+      case "call":
+        return <VideoCameraGray />;
```
gabrielsaruhashi, line 49: seems like this icon logic is used in a couple of places.
Would move to a MeetingIcon component that takes the meeting type and an optional classname as a props

src/redux/modules/inPersonVisit.ts
```diff
@@ -0,0 +1,104 @@
...
+  PayloadAction,
+} from "@reduxjs/toolkit";
+import { fetchAuthenticated } from "src/api/Common";
+import i18n from "src/i18n/config";
+import { Meeting, MeetingStatus } from "src/types/Meeting";
+import { Contact } from "src/types/User";
+import { GeneralStatus } from "src/types/Visit";
+import { notEmpty, showToast } from "src/utils";
+
+const FETCH_IN_PERSON_VISITS = "inPersonVisits/fetchAll";
```
gabrielsaruhashi, line 15: looking at the use cases for meetings on the pathways client, it seems to be we could just have a MeetingsSlice that has both IPV and the calls

It seems like in a couple of places of the codebase we're having to manually merge the two entities back together.

src/hooks/useMeetings.ts
```diff
@@ -0,0 +1,56 @@
...
+  selectAllCalls,
+  selectAllInPersonVisits,
+  selectContactEntities,
+} from "src/redux/selectors";
+import { useAppSelector } from "src/redux";
+import { loadAllInPersonVisitEntities } from "src/redux/modules/inPersonVisit";
+import { Meeting } from "src/types/Meeting";
+
+export function useUpcomingMeetings() {
+  const [calls, setCalls] = useState<Meeting[]>([]);
```
gabrielsaruhashi, line 13: see comment below! 
But I think we should merge the IPV and calls slice into a meetings slice given the use cases that we have

src/components/Dashboard/MeetingsList.tsx
gabrielsaruhashi: in this case, it seems to be just Meeting[]

gabrielsaruhashi: could we call this uniqueMeetingDates?
Think it'd make it more clear what it does

src/components/Dashboard/MeetingsList.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+      const dayOfMonth = new Date(scheduledStart).getDate();
+      const meetingsAtDate = filteredMeetings.filter(
+        (meeting) =>
+          new Date(meeting.scheduledStart).getDay() ===
+          new Date(scheduledStart).getDay()
+      );
+      return (
+        <Row justify="start">
+          <Col flex="50px">
```
gabrielsaruhashi, line 89: we're going to be using tablets real soon
to make this responsive, I'd use flex integers rather than hardcoding the pxs

src/components/Dashboard/MeetingsList.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+    return meetingsByDate.map((meeting) => {
+      const startDate = new Date(meeting.scheduledStart);
+      const date = `${new Date(meeting.scheduledStart)}, ${startDate.getDay()}`;
+      const values = date.split(",");
+      const scheduledStart = values[0];
+      const weekDay = weekDays[parseInt(values[1])];
+
+      const dayOfMonth = new Date(scheduledStart).getDate();
+      const meetingsAtDate = filteredMeetings.filter(
```
gabrielsaruhashi, line 82: it seems like you're first getting the unique dates, then having to re-group all the meetings that happen at that date

Wouldn't it make more sense to just use groupby right away and avoid duplicating the effort to map over the array multiple times?

gabrielsaruhashi: there's a groupBy implementation in Common.ts that you can use

src/components/Dashboard/MeetingsList.tsx
gabrielsaruhashi: I think if you cast it back to string antd will be happy with it
That's what I've done in the past to keep the type safety check

gabrielsaruhashi (changes requested):

src/components/Dashboard/MeetingItem.tsx
lspdrz: Ok, I've added the tailwind styles but we still need to add the left border color like above. The tailwind classNames aren't working with the antd Card border default styles.

Do you have an example or idea of adding spacing on the parent div, I don't follow?

src/components/Dashboard/MeetingsList.tsx
lspdrz: Sure, but why? It'll add another component to save 10 lines of code. Style choice, i guess. What should we call it? MeetingItemWrapper?

src/components/Meetings/Meetings.tsx
lspdrz: Contacts: there are duplicate contacts because that loops through all of the inmate's meetings, so there may be the same contact on multiple meetings

CustomFilter: I don't know, this was copied from Mark's types. Maybe he knows?

cancelnPersonVisit: legacy design, but they use separate language. Let's sort out in another PR? IT's working now

src/components/Meetings/Meetings.tsx
lspdrz: It's only being used in the MeetingTable. The one in the MeetingItem uses different icons

src/redux/modules/inPersonVisit.ts
lspdrz: Yep, agreed. I think we can create a /meetings endpoint that returns both calls and ipvs and use that for pathways

src/hooks/useMeetings.ts
lspdrz: I agree but I think we should do in a new story

src/components/Dashboard/MeetingsList.tsx
lspdrz: I can't find that in Common.ts

src/components/Meetings/Meetings.tsx
gabrielsaruhashi: have u tried without it? I dont think it's needed

src/components/Dashboard/MeetingsList.tsx
lspdrz: I don't know what you mean here

src/components/Meetings/Meetings.tsx
lspdrz: Because this loops through all of the inmate meetings, there might be the same contact on multiple meetings

src/components/Meetings/Meetings.tsx
lspdrz: It's for the styling, I think. I can try it without

src/components/Modals/Modals.tsx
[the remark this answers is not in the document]
lspdrz: Job for a new story

src/types/Meeting.ts
lspdrz: I don't know, I copied this over from mark's types

src/components/Dashboard/MeetingItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const { t } = useTranslation("dashboard");
+
+  const duration = differenceInMinutes(
+    new Date(meeting.scheduledEnd),
+    new Date(meeting.scheduledStart)
+  );
+
+  const canJoin = new Date(meeting.scheduledStart) <= new Date();
+
+  const getIcon = () => {
```
gabrielsaruhashi, line 33: could we move this to a component? MeetingIcon
It's gonna be used in other places

src/components/Dashboard/MeetingsList.tsx
gabrielsaruhashi: you can cast a variable in TS if you're confident that it's what you expect
literally:
activeMeetingTab as string

src/components/Dashboard/MeetingsList.tsx
```diff
@@ -0,0 +1,123 @@
...
+      const startDate = new Date(scheduledStart);
+      const date = `${new Date(scheduledStart)}, ${startDate.getDay()}`;
+      const values = date.split(",");
+      const scheduledStartDate = values[0];
+      const weekDay = weekDays[parseInt(values[1])];
+
+      const dayOfMonth = new Date(scheduledStartDate).getDate();
+      meetingItems.push(
+        <Row justify="start">
+          <Col flex={0.1}>
```
gabrielsaruhashi, line 81: trying to understand what's the intended behavior here - is the idea that it simply occupies the minimum amount of space needed? 
If yes I'd do flex="none"

src/components/Dashboard/MeetingItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+        return <VideoCamera className="mb-2" />;
+    }
+  };
+
+  const getLeftBorderColor = () => {
+    switch (meeting.meetingType) {
+      case "call":
+        return "#569FE1";
+      case "in_person_contact":
+        return "#F66262";
```
gabrielsaruhashi, line 51: bumping this request
cool if we do it in a fast-follow PR, but think it's important to update the palette

src/components/Meetings/Chip.tsx
```diff
@@ -0,0 +1,36 @@
...
+const Chip: React.FC<Props> = ({
+  meetingTypeFilter,
+  setMeetingTypeFilter,
+  meetingType,
+}) => {
+  return (
+    <>
+      <Button
+        type="text"
+        style={{ paddingLeft: "0px" }}
```
gabrielsaruhashi, line 22: is this overriding the antd css?

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,180 @@
...
+        <Row>
+          <Chip
+            meetingTypeFilter={meetingTypeFilter}
+            setMeetingTypeFilter={setMeetingTypeFilter}
+            meetingType="in_person_contact"
+          />
+        </Row>
+      </Menu.Item>
+      <Menu.Item>
+        <Row>
```
gabrielsaruhashi, line 60: nit: i dont think you need to wrap it in a Row here

src/components/Meetings/Meetings.tsx
```diff
@@ -0,0 +1,180 @@
...
+            setMeetingTypeFilter={setMeetingTypeFilter}
+            meetingType="in_person_no_contact"
+          />
+        </Row>
+      </Menu.Item>
+    </Menu>
+  );
+
+  const contactTypesMenu = () => {
+    let contacts = [];
```
gabrielsaruhashi, line 72: @lspdrz why not just pull all contacts directly from the contacts slice? 
It might be off-putting to not see all your approved contacts on your list even if you dont have any scheduled meetings with them. It might seem you dont have the ability to scheduled meetings with them

src/components/Meetings/MeetingsTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+const MeetingsTable: React.FC<Props> = ({
+  meetings,
+  meetingTypeFilter,
+  selectedContactsFilter,
+  setSelectedMeeting,
+}) => {
+  const dispatch = useAppDispatch();
+  const { t } = useTranslation("calls");
+
+  const getIcon = (visitType: MeetingType) => {
```
gabrielsaruhashi, line 30: ^ bumping request to move it to a MeetingComponent

src/components/Meetings/MeetingsTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+        return <PeopleGray />;
+      case "in_person_no_contact":
+        return <PeopleGray />;
+      default:
+        return <VideoCameraGray />;
+    }
+  };
+
+  const filterVisits = () => {
+    return meetings.filter((meeting) => {
```
gabrielsaruhashi, line 44: .filter always returns a new object, so this will keep being re-computed

I'd either create a filterdVisits useState() and update it using a useEffect (preferably) or memoize this function

tailwind.config.js
```diff
@@ -1,8 +1,38 @@
...
+        100: "#EBF4FB",
+        200: "#C8E2F6",
+        300: "#83C5F5",
+        400: "#569FE1",
+        500: "#3577DA",
+        600: "#2E47BD",
+        700: "#0C2C5A",
+      },
+      gray: {
+        // 50: '#fafafa',
```
gabrielsaruhashi, line 20: delete?

gabrielsaruhashi (changes requested):
last round of requests!

gabrielsaruhashi (approved):

src/components/Dashboard/MeetingsList.tsx
lspdrz: flex="none" doesn't work. I'm trying to get a certain amount of distance between the columns: image

src/components/Dashboard/MeetingsList.tsx
lspdrz: Not sure what we want here?

lspdrz: OH, i see, i see

src/components/Dashboard/MeetingItem.tsx
lspdrz: I responded to this elsewhere, but we have to do it this way herre because if I do this style via tailwind the antd component overrides it.

src/components/Meetings/Chip.tsx
lspdrz: Yes

src/components/Meetings/Meetings.tsx
lspdrz: This is a new story, changes multiple files.

src/components/Meetings/MeetingsTable.tsx
lspdrz: What do you mean MeetingComponent?

src/components/Modals/Modals.tsx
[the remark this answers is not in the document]
lspdrz: https://app.clubhouse.io/ameelio/story/304/change-cancelmeetingmodal-on-pathways-to-use-same-dispatch-functions

src/components/Meetings/Meetings.tsx
lspdrz: https://app.clubhouse.io/ameelio/story/305/change-meetings-contact-fetching

src/hooks/useMeetings.ts
lspdrz: https://app.clubhouse.io/ameelio/story/306/move-calls-and-ipv-slices-into-one-meetings-slice

PR #57, opened 2021-08-02
stop assuming that producer update from mobile is totally reliable

gabrielsaruhashi:
## Proposed changes

The producer update event from mobile is flaky. This stops relying heavily on that to generate the UI


mpekala23 (approved):
looks great! will let you know once i figure out what's going on mobile side

PR #60, opened 2021-09-14
fix "see all" path

cainlevy:
There is still a bug with the sidebar, which does not update to reflect the new location. Our sidebar component does not contain any logic to determine which menu item is selected, so I'm guessing ant is somehow unable to observe our navigation changes.


gabrielsaruhashi:
yep! long-standing bug 
We just need to listen to path changes on the Menu component. it's been on the back burner for a while
Will open ticket on clubhouse

gabrielsaruhashi (approved):

PR #61, opened 2021-09-20
Update GH Action Workflow

gabrielsaruhashi:
## Proposed changes

Learning from past mistakes ❌


.github/workflows/deploy.yml
```diff
@@ -3,13 +3,13 @@ name: Build and Deploy
 on:
   push:
     branches:
-      - master
+      - main
```
cainlevy, line 6: so merging `origin/dev` to `origin/main` triggers a release to staging? 👍

cainlevy (approved):

cainlevy:
we may want to add the deployment process to the readme

.github/workflows/deploy.yml
gabrielsaruhashi: yep!

gabrielsaruhashi: oh wait, dev --> would be production

anything to --> dev would be a staging release

PR #64, opened 2021-09-21
added GH workflow to make sure web app can build

gabrielsaruhashi:
Runs install, tsc and build
[ch555]


gabrielsaruhashi:
@cainlevy gonna go ahead and merge this one as it's the same script as that of the DOC web client

PR #71, opened 2021-10-14
Add k8s deploy for staging and prod

FreedomBen:


deploy/nginx.conf
```diff
@@ -0,0 +1,106 @@
...
+    #    ssl_ciphers "ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-      SHA384";
+    #}
+
+    server {
+        listen 8080 default_server;
+        server_name _;
+        root        /app/src/build/;
+
+        add_header 'Access-Control-Allow-Origin' '*';
+        add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
```
cainlevy, line 85: what does CORS do for us?

deploy/nginx.conf
```diff
@@ -0,0 +1,106 @@
...
+
+        add_header X-Frame-Options "SAMEORIGIN";
+        add_header X-XSS-Protection "1; mode=block";
+        add_header X-Content-Type-Options "nosniff";
```
cainlevy, line 89: should we also add HSTS?

deploy/nginx.conf
```diff
@@ -0,0 +1,106 @@
...
+
+        index index.html;
+
+        charset utf-8;
+        client_max_body_size 6M;
```
cainlevy, line 94: do we handle uploads?

deploy/nginx.conf
```diff
@@ -0,0 +1,106 @@
...
+
+        location / {
+
+        }
+
+        location = /favicon.ico { access_log off; log_not_found off; }
+        location = /robots.txt  { access_log off; log_not_found off; }
+
+        error_page 404 /index.php;
```
cainlevy, line 103: 👀 okay okay something tells me this nginx.conf has history ...

k8s/staging/deploy.yaml
```diff
@@ -0,0 +1,55 @@
...
+    matchLabels:
+      app: pathways-client
+  template:
+    metadata:
+      labels:
+        app: pathways-client
+      namespace: pathways-client-staging
+    spec:
+      containers:
+      - image: registry.digitalocean.com/ameelio-registry/pathways-client:prod
```
cainlevy, line 21: staging?

FreedomBen: d'oh, nice catch!

deploy/nginx.conf
FreedomBen: Nope not for this app, although I always like to be explicit with this option so it's greppable and people in the future can find it.  I forget the default but it's pretty low so it's not uncommon for a new endpoint to start biting people and leaving them wondering where in the several steps along the chain the size limit is being enforced :-)

deploy/nginx.conf
FreedomBen: The TLS cert from nginx here is just between us and cloudflare, so user's will never see it

deploy/nginx.conf
cainlevy: this is only a static file server though, right? we don't expect any preflight requests?

cainlevy: (also `Access-Control-Allow-Origin: '*'` shouldn't be set on any API we deploy)

FreedomBen: I could definitely be wrong, but my understanding was that CORS applied to any destinations that the javascript on the page might try to call.  By default javascript on a page can only call the same domain that served it.  If you want to be able to call a different domain (like connect-api.ameelio.org from a page served at pathways-client.ameelio.org) then you would need to allow that.  Is that wrong?

cainlevy: In that scenario Connect API is responsible for authorizing the pathways client's origin with CORS headers. This nginx server, on the other hand, is only hosting the client assets and shouldn't expect any cross origin requests (especially after disabling framing).

deploy/nginx.conf
cainlevy: As far as I know, HSTS simply moves the HTTP->HTTPS redirect into the browser (after the first visit on a new device). That's enough to cut off a potential MITM attack.

What's the downside? I would be surprised if firewalls have anything to say about it. It's as though someone typed https:// instead of http://.

FreedomBen: The HTTP -> HTTPS part is great with no downside that I know of.  Unfortunately it also disables the user's ability to accept the TLS cert in the event that it doesn't pass (for example if it's been replaced by a proxy's cert), or if the cert expires.  At SimpleNexus we had a cert expire and because we used HSTS, none of us could log in to the web app to update the certificate because Chrome wouldn't let us through.  We ended up having to stand up a VM (which browser hadn't yet seen the site so didn't know it was HSTS) and using that to log in to get the cert renewed.  Then for 48 hours there was absolutely no workaround we could provide our users because their browser wouldn't let them proceed.

deploy/nginx.conf
FreedomBen: Cool, I'll remove it

PR #72, opened 2021-10-15
get calls with graphql (feature flagged)

cainlevy:
Here's a functioning proof of concept for querying calls from GraphQL using my favorite type-safe workflow!

## The `typed-document-node` Workflow

1. write a `MyQuery.graphql` file
2. run `npm run graphql` to compile the query (by talking to a GraphQL server)
3. execute the query using the typed DocumentNode (aka compiled query)
4. use the fully typed result object 😍

## How it Looks

image

Depends on https://github.com/Ameelio/connect-api-nest/pull/358.


codegen.yml
```diff
@@ -0,0 +1,14 @@
+overwrite: true
+schema: http://${API_HOST}:${API_PORT}/api/graphql
+generates:
+  src/api/graphql.ts:
+    plugins:
+      - typescript
```
cainlevy, line 6: this file is generated with basic type information, like a list of enums the server might return

codegen.yml
```diff
@@ -0,0 +1,14 @@
...
+  src/:
+    documents: src/**/*.graphql
+    preset: near-operation-file
+    presetConfig:
+      baseTypesPath: api/graphql.ts
+    plugins:
+      - typescript-operations
+      - typed-document-node
```
cainlevy, line 14: each `MyQuery.graphql` file gets compiled into a sibling `MyQuery.ts` file that you can import

codegen.yml
```diff
@@ -0,0 +1,14 @@
+overwrite: true
+schema: http://${API_HOST}:${API_PORT}/api/graphql
```
cainlevy, line 2: this URL assumes folks are running https://github.com/Ameelio/connect-dev

you can always export ENV vars yourself, though.

package.json
```diff
@@ -36,6 +37,7 @@
     "reselect": "^4.0.0",
     "socket.io-client": "^3.1.2",
     "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.0.3",
+    "urql": "^2.0.5",
```
cainlevy, line 40: i picked urql because it's smaller and more lightweight than apollo. we're dipping our toes in the water here.

src/api/ClientProvider.tsx
```diff
@@ -0,0 +1,19 @@
...
+    return {
+      headers: { authorization: token || "" },
+    };
+  },
+  exchanges: [fetchExchange]
+});
+
+const ClientProvider: React.FC = ({ children }) => (
+  <Provider value={client}>{children}</Provider>
+);
```
cainlevy, line 17: we don't actually use this ClientProvider, because our components use Redux thunks instead of `useQuery`.

src/api/graphql.ts
```diff
@@ -0,0 +1,149 @@
+export type Maybe<T> = T | null;
```
cainlevy, line 1: pretty much always ignore generated files like this

src/redux/index.ts
```diff
@@ -4,7 +4,6 @@ import { History, createHashHistory } from "history";
 import { configureStore } from "@reduxjs/toolkit";
 import { sessionReducer } from "./modules/session";
 import { callSlice } from "./modules/call";
-import { connectionSlice } from "./modules/connection";
```
cainlevy, line 7: i cleaned up some unused types while i tried to figure out what data we actually _needed_ to query

src/redux/modules/GetMeetings.graphql
```diff
@@ -0,0 +1,29 @@
+query GetMeetings {
```
cainlevy, line 1: it's a query!

NOTE: query names must be unique across an entire application because of how our codegen works.

src/redux/modules/call.ts
```diff
@@ -1,34 +1,73 @@
...
+import { GetMeetingsDocument } from "./GetMeetings.generated";
 
 const FETCH_CALLS = "calls/fetchAll";
-export const fetchCalls = createAsyncThunk(FETCH_CALLS, async () => {
-  const body = await fetchAuthenticated(`calls`);
+export const fetchCalls = createAsyncThunk(
+  FETCH_CALLS,
+  async (): Promise<Call[]> => {
+    // GRAPHQL
+    if (featureEnabled("GRAPHQL")) {
```
cainlevy, line 33: our server doesn't run the GraphQL endpoint in production, so this experiment has to be limited to local environments that opt in with a `REACT_APP_ENABLE_GRAPHQL=1` var.

src/redux/modules/call.ts
```diff
@@ -1,34 +1,73 @@
...
+        userParticipants: call.visitors.map((v) => ({
+          ...v,
+          status: "active", // HAX!
+        })),
+        videoHandler: call.call?.url
+          ? {
+              url: call.call.url || "",
+            }
+          : undefined,
+      }));
```
cainlevy, line 62: type-safe response formatting (except for the `... as MeetingStatus`).

this typically shouldn't be necessary, but i didn't want to refactor the app in any way. it'd be easier after all of the thunks are switched over in production, anyway.

src/redux/modules/call.ts
```diff
@@ -1,34 +1,73 @@
...
-  const calls = (body.data as { results: Call[] }).results;
+      // slice out calls
+      const calls = result.data.meetings.edges
+        .map(({ node }) => node)
+        .filter((meeting) => meeting.meetingType === MeetingType.Call);
 
-  return calls;
-});
+      // convert to legacy data types
+      return calls.map((call) => ({
```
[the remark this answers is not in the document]
cainlevy: yeah the transition to graphql is a good time to revisit some of our api and model choices. let's take advantage of it!

src/redux/modules/call.ts
```diff
@@ -1,34 +1,73 @@
...
+        id: call.id,
+        meetingType: "call",
+        status: call.status.toLowerCase() as MeetingStatus, // HAX!
+        kioskName: call.kiosk.name,
+        scheduledStart: call.interval.startAt,
+        scheduledEnd: call.interval.endAt,
+        userIds: call.visitors.map((v) => v.id),
+        userParticipants: call.visitors.map((v) => ({
+          ...v,
+          status: "active", // HAX!
```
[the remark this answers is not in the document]
cainlevy: yeah, that's my understanding. i set it to "active" because it seems fair to assume when the server just told us they're still a meeting participant.

src/utils/featureEnabled.ts
```diff
@@ -0,0 +1,3 @@
+export default function featureEnabled(name: string): boolean {
+  return !!process.env[`REACT_APP_ENABLE_${name}`];
+}
```
gabrielsaruhashi, line 3: do we want to make "name" an enum to be more strictly enforced? / so that everyone knows all the flags that exist

cainlevy: oh that sounds pretty nice

PR #82, opened 2021-11-02
Reorganized call feedback page

gabrielsaruhashi:
## Proposed changes
Reorganized call feedback page
https://user-images.githubusercontent.com/12942772/139911189-2136f4f6-d9e1-4828-8a23-e840e3d676c9.mov

[sc-1351]


src/components/Typography/index.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+
+interface Props {
+  className?: string;
+  children?: React.ReactNode | React.ReactNode[];
+  bold?: boolean;
+}
+
+export const H1 = ({ className, children, bold }: Props) => {
```
gabrielsaruhashi, line 9: figured it was a good time to add this typography layer
will make it easier for us to make the app responsive to different tablet sizes in the future

src/pages/CallFeedback/index.tsx
```diff
@@ -1,60 +1,36 @@
...
 import Loader from "src/components/Loader";
 import { logout } from "src/redux/modules/session";
 import { addMinutes } from "date-fns";
 import { rateCall } from "src/api/Call";
-import { selectAllCalls } from "src/redux/selectors";
 import { updateProfile } from "src/api/User";
+import { useIntakeForm } from "src/hooks/useIntakeForm";
 
-type TParams = { id: string };
+type TParams = { id: string; exitType: CallExitType };
```
gabrielsaruhashi, line 19: this is setup work for the next PR where we start tracking whether the other participant ever attempt to join.

The call page will set the url param according to the state of RoomClient right before the user leaves the room.

The alternative was to re-load the room client in this page, but that felt overkill for this use case

src/components/CallFeedback/CallFeedback.tsx
```diff
@@ -1,282 +1,157 @@
...
-  markIntakeFormAsComplete: () => void;
+  onSubmitIntakeForm: () => void;
+  promptIntakeForm: boolean;
 }
 
+const FeedbackWrapper = ({
+  children,
+}: {
+  children: React.ReactNode[] | React.ReactNode;
+}) => (
```
cainlevy, line 42: the `React.FC` type will take care of declaring children if you like:

```suggestion
const FeedbackWrapper: React.FC = ({children}) => (
```

src/components/Typography/index.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+
+interface Props {
+  className?: string;
+  children?: React.ReactNode | React.ReactNode[];
```
cainlevy, line 5: i recommend React.FC

cainlevy (approved):
i didn't spend the time to figure out what was new vs refactored, so i just skimmed for obvious lint.

PR #84, opened 2021-11-02
only prompt participant to rate if the other party ever attempted to join

gabrielsaruhashi:
@cainlevy still debugging my kind environment, but the logic is pretty simple, so ill just go ahead ask for your review


src/pages/Call/index.tsx
```diff
@@ -129,9 +129,16 @@ const CallBase: React.FC<RouteComponentProps<TParams>> = ({ match }) => {
...
-    () => dispatch(push(`/feedback/${callId}`)),
-    [callId, dispatch]
+  const leaveCall = useCallback(
+    (exitType: CallExitType) => {
+      if (!call) {
+        dispatch(push("/"));
+        return;
+      }
+
+      dispatch(push(`/feedback/${call.id}?exitType=${exitType}`));
```
cainlevy, line 139: the routing below makes me think this should be a path?

```suggestion
      dispatch(push(`/feedback/${call.id}/${exitType}`));
```

cainlevy (approved):

PR #88, opened 2021-11-03
Skip dev login

gabrielsaruhashi:
## Proposed changes
Makes me happy.


src/pages/Login/index.tsx
```diff
@@ -113,7 +113,9 @@ const LoginContainer: React.FC = () => {
             onFinish={onFinish}
             initialValues={{
               language: Object.keys(LANGUAGES)[0],
-              pin: "",
+              pin: process.env.NODE_ENV === "development" ? "inmatemember" : "",
+              inmateNumber:
+                process.env.NODE_ENV === "development" ? "Ameelio-001" : "",
```
cainlevy, line 118: i believe these values will persist in our production code, which means they'll be visible to anyone who decides to inspect the source.

maybe we could have specific ENV vars for this? that would also let me specify `Ameelio-002` when they happen to have the call i'm debugging. 😄

gabrielsaruhashi: yep!

cainlevy (approved):

PR #92, opened 2021-11-04
WIP gql GetContacts query

wittejm:
Used the edge/nodes structure like Meetings uses. But I notice @gabrielsaruhashi  just added Facilities without adding this; is that because facilities query results will be smaller and less likely to need pagination?


src/api/graphql.ts
```diff
@@ -130,7 +155,9 @@ export type Query = {
   me?: Maybe<User>;
   meeting: Meeting;
   meetings: MeetingConnection;
+  contacts: ContactConnection;
```
cainlevy, line 158: > is that because facilities query results will be smaller and less likely to need pagination

yep! i chatted with @gabrielsaruhashi about it briefly. he says there are 5k facilities in the US which is definitely enough for pagination. we just can't imagine (yet) any use cases where a page would want to list all of them. the current use case (pathways login) would need to be rethought.

deciding to use a connection (or not) is a gamble. it can save us work in the future but it creates work now.

src/redux/modules/GetContacts.graphql
```diff
@@ -0,0 +1,18 @@
+query GetContacts {
+  contacts {
```
cainlevy, line 2: @gabrielsaruhashi you observed a while back that we're starting to use "connection" for more things. any opinions about "contact" as an alternative? if we like it, we can start by adopting it in the API.

src/redux/modules/GetContacts.graphql
```diff
@@ -0,0 +1,18 @@
...
+        id
+        inmateIdentification
+        dateOfBirth
+        callQuota
+        visitQuota
+        location
+        race
+        needsApproval
+        needsMonitor
+        hasFilledIntakeForm
```
cainlevy, line 14: how many of these are used by the pathways client currently?

wittejm: Oh woops, wrong fields. I was pulling from the wrong section of https://github.com/Ameelio/pathways-client/blob/main/src/types/User.ts
sloppy!

src/redux/modules/GetContacts.graphql
gabrielsaruhashi: oop! sorry I just noticed this
I believe contact might not make it clear that it's a relationship between two people?
How about "relation"?

src/api/graphql.ts
gabrielsaruhashi: For preciseness:
>This report offers some much needed clarity by piecing together this country’s disparate systems of confinement. The American criminal justice system holds almost 2.3 million people in 1,833 state prisons, 110 federal prisons, 1,772 juvenile correctional facilities, 3,134 local jails, 218 immigration detention facilities, and 80 Indian Country jails as well as in military prisons, civil commitment centers, state psychiatric hospitals, and prisons in the U.S. territories. This report provides a detailed look at where and why people are locked up in the U.S., and dispels some modern myths to focus attention on the real drivers of mass incarceration, including exceedingly punitive responses to even the most minor offenses.


I'd say that for pathways, we def don't need to paginate contacts in my opinion. Judy is one of our most "popular" users and she has 10 contacts.

We probably want to do it for the facilities tho.

src/redux/modules/GetContacts.graphql
cainlevy: oof "relation" is also pretty overloaded in data modeling.

i mean, sometimes that's just how it goes. at my last gig we had FooModel models.

PR #93, opened 2021-11-08
query meetings through me{...}

cainlevy:
Fixing Pathways after https://github.com/Ameelio/connect-api-nest/pull/439.


src/redux/modules/GetMeetings.graphql
```diff
@@ -1,27 +1,31 @@
...
-          firstName
-          lastName
-          profileImageUrl
-        }
-        call {
-          status
-          url
-          token
+  me {
+    ... on Inmate {
```
cainlevy, line 3: this is a little awkward. i may follow up with an idea from last week's tour with @wittejm.

src/redux/modules/call.ts
```diff
@@ -32,12 +32,16 @@ export const fetchCalls = createAsyncThunk(
     // GRAPHQL
     if (featureEnabled("GRAPHQL")) {
       const result = await client.query(GetMeetingsDocument).toPromise();
-      if (result.error || !result.data) {
+      if (
+        result.error ||
+        !result.data ||
+        result.data.me?.__typename !== "Inmate"
```
cainlevy, line 38: also awkward. this helps typescript!

src/redux/modules/GetMeetings.graphql
gabrielsaruhashi: is this the syntax to do conditional loading?

gabrielsaruhashi (approved):

src/redux/modules/GetMeetings.graphql
cainlevy: yep!

we're going to get friendly with it in a timeline query.

PR #95, opened 2021-11-09
graphql: cancelMeeting

cainlevy:


gabrielsaruhashi (approved):

src/api/graphql.ts
```diff
@@ -36,7 +36,7 @@ export type CancelMeetingInput = {
 
 export type CancelMeetingResponse = {
   __typename?: "CancelMeetingResponse";
-  meeting?: Maybe<Meeting>;
+  meeting: Meeting;
```
gabrielsaruhashi, line 39: oh why is this optional?

cainlevy: it used to be. fixed in https://github.com/Ameelio/connect-api-nest/pull/460

PR #96, opened 2021-11-09
hide unfinished cancellation reason

cainlevy:
Pathways has been collecting a cancellation reason that our NestJS API has ignored. The idea was that we could send the reason along with the SMS/email notification, but it was never implemented and would need to be revisited to disallow the possibility of free text communication.

Should we remove the dropdown from this modal until we can focus on better support?

image


gabrielsaruhashi (approved):
cool with this. 
is the solution to keep the string representation on the underlying model, but create a string enum on the Graphql mutation?

If yes, I'd be happy to create two tickets to get this feature back up

cainlevy:
@gabrielsaruhashi yep sounds about right. Should I close this and assume it'll be hooked up soon?

cainlevy:
I'll put it behind a feature flag 😄

PR #97, opened 2021-11-09
remove messages redux logic

cainlevy:
We don't plan to implement messages with redux. Might as well remove it until we're ready to build?


gabrielsaruhashi (approved):

gabrielsaruhashi:
rip my nest micro-service experiment

PR #109, opened 2021-11-11
fix an infinite re-render

cainlevy:
1. `Call.tsx` calls `updateCallStatus` when `onLeave` changes
2. when `updateCallStatus` is invoked, redux changes the `call` (even on a no-op)
3. `onLeave` was recomputed each time `call` changes

The simplest fix is to avoid a closure with onLeave (3). The deeper problem appears to be mutating our API object with UI state (2).


FreedomBen (commented):
Hahaha, it's amazing sometimes how small the patches are to fix crazy issues :laughing:

PR #115, opened 2021-11-29
send message mutation

gabrielsaruhashi:
## Proposed changes
Sends the message and updates the cache.
Ignore the yellow "focus" background on the demo :) . That has been removed

https://user-images.githubusercontent.com/12942772/143912212-89f9d87b-0c68-4d5f-8e6a-2e7cc1116bc3.mov


src/components/Inbox/MessageCompose.tsx
```diff
@@ -4,8 +4,9 @@ import { useState } from "react";
 import { useTranslation } from "react-i18next";
 
 interface Props {
-  onSend: (content: string) => void;
+  onSend: (content: string) => Promise<void>;
   className?: string;
+  loading?: boolean;
```
cainlevy, line 9: is loading used somewhere?

src/components/Inbox/MessageList.tsx
```diff
@@ -18,18 +17,18 @@ const MessagesList = ({ messages, focus }: Props) => {
       behavior: "smooth",
       block: "start",
     });
-  }, [focusRef, focus, messages]);
+  }, [focusRef, messages]);
```
cainlevy, line 20: if this `useEffect` is working properly, then i have to assume React runs it after updating `focusRef`. like, maybe react updates refs _during_ rendering and runs effects _after_ rendering.

src/pages/InboxPage.tsx
```diff
@@ -17,11 +19,43 @@ const InboxPage: React.FC<Props> = () => {
...
   const connections = connectionData?.currentInmate?.connections || [];
+
   const messages =
     messagesData?.connection.messages.edges.map(({ node }) => node) || [];
+  const [createMessage, { loading: createLoading, error: createError }] =
+    useMutation(CreateMessageDocument, {
+      update: (cache, { data }) => {
+        const newMessage = data?.createMessage.message;
+        if (!newMessage) return;
+        cache.updateQuery(
```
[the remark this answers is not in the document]
cainlevy: happy to help debug. i presume you were looking at https://github.com/Ameelio/connect-doc-client/pull/171/files#diff-a31867d783145cb05c11f8e76e954785362f0793829eefb0efa35c795096a122?

src/pages/InboxPage.tsx
```diff
@@ -17,10 +19,42 @@ const InboxPage: React.FC<Props> = () => {
...
+            };
+          }
+        );
+      },
+    });
+
+  useEffect(() => {
+    if (!createError) return;
+    showToast("createError/inbox", "Failed to send message", "error");
+  }, [createError]);
```
cainlevy, line 57: sentry?

cainlevy (changes requested):

src/pages/InboxPage.tsx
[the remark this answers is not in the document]
gabrielsaruhashi: yep!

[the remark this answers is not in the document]
gabrielsaruhashi: Is updateQuery incorrect?
It seems to accomplish the same as modify, but has less options/control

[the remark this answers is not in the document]
cainlevy: `updateQuery` and `cache.modify` appear to accomplish roughly the same thing. `cache.modify` is lower-level and lets us operate on the normalized cache without the query abstraction.

i have found `cache.modify` to be more versatile because it lets me create decoupled form components. maybe it could work for us here, if we are very careful to colocate `useQuery` and `useMutation` hooks in our page components?

src/components/Inbox/MessageCompose.tsx
gabrielsaruhashi: oop now it is!

src/components/Inbox/MessageList.tsx
gabrielsaruhashi: actually, i realized I can actually simplify this a lot :) 
the use case was a bit different for DOC

src/utils/Message.ts
```diff
@@ -17,9 +18,8 @@ export function getMessageStatusLabel(status: MessageStatus) {
...
 
 export function getMessageSenderType(
-  message: TMessageItem
+  message: TMessageItem,
+  connection: TConnectionListItem
 ): "inmate" | "visitor" {
-  return message.senderId === message.connection.visitor.malanId
-    ? "visitor"
-    : "inmate";
+  return message.senderId === connection.visitor.malanId ? "visitor" : "inmate";
```
gabrielsaruhashi, line 24: will move this computation to the server Resolver v soon :)

cainlevy, line 24: i see why it felt right to have `message.connection.visitor` previously.

what we really want is `message.sender: Visitor | Inmate`. is that part of the `malanId` refactor we talked about previously?

cainlevy (approved):

src/utils/Message.ts
gabrielsaruhashi: yep!

PR #117, opened 2021-12-04
Update unread messages as read upon MessageList load

wittejm:
This doesn't immediately modify the cache with the new isRead value, because we need it to show the "new" dot next to each message. I'm trying to get clever with the exact right time to modify the cache, but that's still a WIP.


src/components/Inbox/MessageList.tsx
```diff
@@ -21,6 +28,16 @@ const MessagesList = ({ messages, connection }: Props) => {
     });
   }, [focusRef, messages]);
 
+  useEffect(() => {
+    messages.filter(m => !m.isRead).map(m => updateMessageIsRead({ variables: { input: { messageId: m.id } } }))
+  }, [messages, updateMessageIsRead]);
+
+  useEffect(() => {
+    if (!updateIsReadError) return;
+    showToast("updateIsRead", "Failed to update message is-read status ", "error");
```
cainlevy, line 37: i'd be in favor of not toasting this one. doesn't feel important enough.

cainlevy (approved):

PR #119, opened 2021-12-09
adds emessaging flag to graphql query

lspdrz:
[sc-1632]
DO NOT MERGE UNTIL https://github.com/Ameelio/connect-api-nest/pull/557 is merged and deployed

Adds a FacilityServices fragment that can be used in any request that contains a service hidden behind a feature flag


src/hooks/useFacilityFeature.ts
```diff
@@ -0,0 +1,19 @@
+import { useQuery } from "@apollo/client";
+import { GetFacilityServicesDocument } from "src/graphql/queries/GetFacilityServices.generated";
+
+const features = ["MESSAGING"] as const;
+type Feature = typeof features[number];
+
+const useFacilityFeature = (feature: Feature) => {
+  const { data: facilityServicesData } = useQuery(GetFacilityServicesDocument);
+  const facility = facilityServicesData?.currentInmate?.facility;
```
gabrielsaruhashi, line 9: this works for now! in the near future, we'll probably have to change this to return an object

Record<Feature, boolean> so we don't need to do multiple calls

gabrielsaruhashi (approved):

src/components/Sidebar/Sidebar.tsx
```diff
@@ -65,7 +65,7 @@ export default function Sidebar({ user, logout }: Props): ReactElement {
             {t("meetings")}
           </Menu.Item>
           {/* TODO: https://app.shortcut.com/ameelio/story/1632/hide-pathways-features-behind-a-facility-based-feature-flag-depending-on-whether-or-not-facility-offers-the-service */}
-          {featureEnabled("MESSAGING") && (
+          {useFacilityFeature("MESSAGING") && (
```
cainlevy, line 68: oh wow i've never seen a hook used in the render before. i guess it works here!

src/hooks/useFacilityFeature.ts
cainlevy: 🤔 we could also probably set an explicit `fetchPolicy: "cache-first"` on the query to accomplish the same

cainlevy (approved):

PR #120, opened 2021-12-09
Jordanwitte/sc 1721/badge with number of unread messages on connectionlistitem

wittejm:
The main puzzle to figure out was minimizing the triggers for useEffect that update or reset the cache. Using `selectedId` rather than `selected` (message) helped with this.

I realized eventually that the updateMessageIsRead mutation was updating the cache right away.
- Added fetchPolicy:'no-cache' to the mutation to disable this.
- But looking at the network panel I see that after all the MessageCard ops run (including mutation and cache stuff), the two queries in InboxPage would rerun! these are GetConnections and GetConnectionMessages. Rerunning these fetches gets the new value which is isRead=true. The cache is overwritten and we lose our shiny 'new message' badges.
- I added nextFetchPolicy:'cache-only' to those queries. This has the desired effect and the app all works according to design! Hooray!
- But I don't like having that nextFetchPolicy. I'd expect the query shouldn't rerun at all. So I need to know why it's rerunning and see if I can fix it.
- Answer from  [apollographql/apollo-client#6702](here): Apparently that refetch *should* happen as we're updating the content of the query using the mutation
- As discussed there, setting nextFetchPolicy is the proposed solution.
- Everything now works as intended.

https://user-images.githubusercontent.com/2104990/145454344-51d87ddb-f51f-4b4b-b583-f90fe1ffb165.mov


wittejm:
Oh. still a build issue. standby...

cainlevy (approved):
> the updateMessageIsRead mutation was updating the cache right away.

Another fix might be removing the response data from the mutation. You'll still need to resolve _something_, but `__typename` is always available:

```
mutation MyMutation(...) {
  myMutation(...) {
    __typename
  }
}
```

wittejm:
> Another fix might be removing the response data from the mutation. You'll still need to resolve something, but __typename is always available:

Oh nice. I actually tried this at one point and it didn't seem to work... works now though. But at the time I think there were other reasons why the parent query was updating.

PR #121, opened 2021-12-09
Lp new login

lspdrz:
https://user-images.githubusercontent.com/9705113/145461960-670b7791-557a-4424-a2b7-e900e3adbec7.mov

[sc-1384]

DO NOT MERGE UNTIL https://github.com/Ameelio/connect-api-nest/pull/559 is merged and deployed

Changes the login page to include state selection


src/utils/enums/SupportedUSState.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum SupportedUSState {
+    'CO' = 'Colorado', // should I add later?
```
lspdrz, line 2: @gabrielsaruhashi should we add Colorado to the dropdown now or later?

src/pages/Login/index.tsx
```diff
@@ -81,15 +84,25 @@ const LoginContainer: React.FC = () => {
...
   };
 
-  const facilityOptions = facilities.map((facility: Facility) => (
-    <Option key={`option-${facility.id}`} value={facility.id}>
-      {facility.name}
-    </Option>
-  ));
+  const stateOptions = Object.keys(SupportedUSState).map((supportedState) => (
+        <Option key={`option-${supportedState}`} value={supportedState}>
+          {SupportedUSState[supportedState as keyof typeof SupportedUSState]}
```
cainlevy, line 89: might be able to bypass this type mismatch with `Object.entries`

src/pages/Login/index.tsx
```diff
@@ -145,16 +160,31 @@ const LoginContainer: React.FC = () => {
...
+              label={t("placeholder.state")}
+              labelAlign={"left"}
+              rules={[{ required: true, message: "Must select a state." }]}
+            >
+              <Select onSelect={(value: UsState) => setState(value)}>{stateOptions}</Select>
+            </Form.Item>
+
+
+{
+  state && (
```
cainlevy, line 174: is prettier working?

wittejm (approved):

src/utils/enums/SupportedUSState.enum.ts
cainlevy: can we get the list of states for the dropdown from the fetched facility data? that would prevent empty facility dropdowns, too.

we'd still need state names, presumably, but that could come from i18n or some other static file.

src/pages/Login/index.tsx
lspdrz: Yep good idea. Or object.values works too turns out

src/pages/Login/index.tsx
lspdrz: now it is!!

src/utils/enums/SupportedUSState.enum.ts
lspdrz: Changed!!1 And yeah we still need the state names, I just used the enum for now

PR #122, opened 2021-12-10
adds profile info drawer to message page

lspdrz:
[sc-1720]
DO NOT MERGE UNTIL https://github.com/Ameelio/connect-api-nest/pull/562 is merged and deployed

image


src/api/graphql.ts
```diff
@@ -56,11 +58,24 @@ export type Connection = {
...
   inmate: Inmate;
   messages: MessageConnection;
+  relationship: ConnectionRelationship;
   restrictions: Array<ConnectionRestriction>;
   status: ConnectionStatus;
   visitor: Visitor;
 };
 
+export enum ConnectionRelationship {
+  Brother = "BROTHER",
```
wittejm, line 68: Do these need to be gendered? We don't have gender on User profiles and these could all be Sibling, Parent, or Child (also Daughter is missing anyway :) )

lspdrz: Ah, daughter is missing bc of a typo on my end. As for the gendered question, we'd need to talk to Liz

src/components/Inbox/InboxHeader.tsx
```diff
@@ -1,21 +1,39 @@
...
+          <div className="flex items-center gap-2">
+            <VisitorAvatar visitor={connection.visitor} size="large" />
+            <span className="font-medium text-lg">
+              {getFullName(connection.visitor)}
+            </span>
+          </div>
+          <InfoCircleOutlined
+            className="cursor-pointer"
+            onClick={() => {setViewInfo(true)}}
+            style={{fontSize: '25px', color: '#8c8c8c'}}
```
wittejm, line 28: Can these be done with Tailwind?

wittejm (approved):
Still missing Daughter from the list. Otherwise LGTM.

src/components/Inbox/InboxHeader.tsx
lspdrz: Unfortunately not, one of those antd things

lspdrz: The fontSize here is actually controlling the width of the icon hahaha

PR #124, opened 2021-12-13
Update deploy script to latest (get_namespace update)

FreedomBen:


scripts/deploy-release.sh
```diff
@@ -357,9 +357,16 @@ get_namespace ()
     manifest=$(deploy_manifest)
   fi
 
-  kubectl $(k8s_server) $(k8s_token) $(k8s_ca) get \
-    -f "${manifest}" \
-    -o jsonpath='{.items[0].metadata.namespace}'
+  #kubectl $(k8s_server) $(k8s_token) $(k8s_ca) get \
+  #  -f "${manifest}" \
+  #  -o jsonpath='{.items[0].metadata.namespace}'
```
cainlevy, line 362: is this still needed?

cainlevy (approved):

scripts/deploy-release.sh
FreedomBen: oh no I don't think so.  I left it in so I could quickly see what the old code did in case one of the projects didn't behave properly, but at this point the new code is way more robust, so I can take that out

PR #126, opened 2021-12-14
First message sent modal

wittejm:
https://user-images.githubusercontent.com/2104990/145909038-14fcb34c-670d-479a-b524-a64ad2150bf7.mov

The infinite loop issue I mentioned in standup was because this query was colliding in the cache on the currentInmate object; commit 2 fixes this by adding `id` to the query:
https://github.com/Ameelio/pathways-client/blob/f9e42970da7a97cd4031abd9bfc257221c191f6d/src/hooks/useFacilityFeature.ts#L8 

Commit 3 enables merge on `connection.messages` because two different queries fetch connection.messages.nodes and connection.messages.unreadMessageCount, respectively. Following this documentation: https://www.apollographql.com/docs/react/caching/cache-field-behavior/ . If I'd realized this earlier it could have solved some of my headaches from last week.  

Commits 4 and 5 make sure the two modals only ever open once. Evaluating a boolean for three states `true` or `false` or `undefined` totally works but is totally a bad idea! 

I set the modal width to 570. I know we generally consider pixel styles not great but 
a) the antd modal width is a hard-coded 520
b) without the slightly expanded width, the text wraps around and looks terrible:

modal_with_wraparound

If there's a more tailwindy way to bump the content width a little, I'm happy to do that instead.


lspdrz (approved):
Nice catch!! I ran into a similar caching issue the other day, where the id was missing in my query so apollo couldn't find the object. Think it's good practice to just always include the id for objects

PR #128, opened 2021-12-15
fixed login page css

lspdrz:
[sc-1877]

image
image


tanyabrassie (approved):

gabrielsaruhashi:
@lspdrz could we merge this?
seems stale

lspdrz:
> @lspdrz could we merge this? seems stale

it never got the green light from Liz, but i think it's good. merging

PR #131, opened 2021-12-16
integrate useConnectCall from connect-call-client

cainlevy:
I've extracted, refactored, and thoroughly tested our CVH integration in this new public repository: https://github.com/Ameelio/connect-call-client

This refactors Pathways to use the shared lib and clarifies the responsibilities of the main components. It will also make refactoring to `useQuery` much easier.

1. `src/pages/Call` is now primarily responsible for integrating with the Connect API. It avoids UI work.
2. `src/components/Call` is now primarily responsible for providing the user interface. It relies on `useConnectCall()` for CVH integration.


src/components/Call/Call.tsx
```diff
@@ -47,160 +40,155 @@ declare global {
...
+      call: {
+        id: call.id,
+        url: call.videoHandler!.url,
+        token: call.videoHandler!.token,
+      },
+      authInfo,
+      onPeerConnected: handlePeerConnected,
+      onPeerDisconnected: handlePeerDisconnected,
+      onNewMessage: handleNewMessage,
+    });
```
cainlevy, line 127: this is the React API of the new CVH client. it supports a few event listeners and emits useful pre-packaged data. the rest of this component is responsible for connecting that data to the user interface.

src/pages/Call/index.tsx
```diff
@@ -1,282 +1,64 @@
...
+ * CallPage handles the call route and is responsible for finding and validated
+ * the requested call record, like any other page. It does not understand how
+ * call rooms work.
+ */
+const CallPage: React.FC<RouteComponentProps<TParams>> = ({ match }) => {
   const dispatch = useAppDispatch();
+  const { authInfo, user } = useAppSelector((state) => state.session);
 
   const callId = match.params.id;
   const call = useCallById(callId);
```
cainlevy, line 30: the next step in a pathways graphql refactor is trading out this redux selector for a `useQuery`.

src/components/Call/Call.tsx
gabrielsaruhashi: this is so incredible!
Over the past months, I always wondered there must have been a good react hook version of all the setup steps - now I learn :D

src/components/Call/Call.tsx
```diff
@@ -47,160 +40,155 @@ declare global {
...
-        return t("waitingRoom.initialization");
-      } else if (participantJoinStatus === "no_show") {
-        return `${t("waitingRoom.waitingForPrefix")} ${getContactsFirstNames(
-          call.userParticipants
-        )} ${t("waitingRoom.waitingForSuffix")}...`;
-      }
-      return t("waitingRoom.loading");
-    };
+    // alert when remote video pauses or resumes
+    const remoteVideoTracks = peer?.stream.getVideoTracks().length;
```
gabrielsaruhashi, line 164: oh this is so genius to avoid infinite re-rendering. 
I always thought this would require a useMemo
I guess this does assume that the video tracks themselves within the array are not gonna mutate over time, which in this case makes a lot of sense

gabrielsaruhashi (approved):
alt

PR #135, opened 2021-12-20
Message statuses hifi

wittejm:
Changes:
- Don't show "delivered" or checkmark on the visitor's messages (it's redundant info).
- Location of status icon is different for rejected vs pending/delivered.
- Explanatory text for unavailable messages.
- Remove sender name or "Me" from every message. 

six_statuses_demo


src/components/Inbox/MessageCard.tsx
```diff
@@ -14,41 +14,62 @@ import { UpdateMessageIsReadDocument } from "src/graphql/mutations/UpdateMessage
...
+        return <span>{message.content}</span>;
+      case MessageAvailability.Pending:
+        return senderType === "inmate" ? <span>{message.content}</span> : <span className="text-gray-400">This message is hidden to you while pending review.</span>;
+      case MessageAvailability.Rejected:
+        return senderType === "inmate" ? <span>{message.content}</span> : <span className="text-gray-400">This message is unavailable because it was rejected.</span>;
       default:
         assertNever(message.availability);
     }
   };
+  const messageAvailabilityText = (() => {
```
wittejm, line 61: building a string const this way with an inline function call seems... better than using a nested ternary, I suppose. I guess I don't like either way. The ternary would be
```suggestion
const messageAvailabilityText = (
  message.availability === MessageAvailability.Available && senderType === "inmate" ?
     "Delivered" : 
     message.availability === MessageAvailability.Pending ?
      "Review Required" : 
      message.availability === MessageAvailability.Rejected ?
        "Rejected" : 
        null
);
```

src/components/Inbox/MessageCard.tsx
```diff
@@ -14,41 +14,62 @@ import { UpdateMessageIsReadDocument } from "src/graphql/mutations/UpdateMessage
...
     switch (message.availability) {
+      case MessageAvailability.Available:
+        return senderType === "inmate" ? <CheckOutlined className="text-blue-500" /> : null;
       case MessageAvailability.Pending:
         return <ClockCircleOutlined className="text-gray-500" />;
       case MessageAvailability.Rejected:
-        return <StopOutlined className="text-red-500" />;
+        return null;
+      default:
+        assertNever(message.availability);
```
gabrielsaruhashi, line 46: this will throw an error right?
why not just return null here?

gabrielsaruhashi: it'd be pretty disruptive for the department of corrections if we update the Graphql type and forget to update the switch before we deploy

src/components/Inbox/MessageCard.tsx
gabrielsaruhashi: def not nested a ternary! it's v hard to read

gabrielsaruhashi: i think this anonymous function implementation works well!

src/components/Inbox/MessageCard.tsx
```diff
@@ -81,35 +102,32 @@ const MessageCard: React.FC<Props> = ({
...
           )}
       </div>
-      <div className="flex gap-1">
-        <Typography.Text type="secondary">{senderName}</Typography.Text>
-      </div>
-      <Typography.Text type="secondary" className="text-xs">
-        {getMessageAvailabilityLabel(message.availability)}
+      <Typography.Text type="secondary" className="text-xs flex items-center">
+        <MessageDeliveryIcon />
+        <span className="pl-1" />
```
gabrielsaruhashi, line 127: what is this empty span for? is this just creating margin?
Why not wrap messageAvailabilityText in this span

gabrielsaruhashi (approved):

src/components/Inbox/MessageCard.tsx
wittejm: Actually this is a type check, so TS will error. Apparently it will _also_ throw a runtime error, which turns out is optional (and now I've disabled that!) but I don't know how that is logically possible anyway, given the earlier TS error.

src/components/Inbox/MessageCard.tsx
wittejm: Yep, makes sense!

PR #137, opened 2021-12-21
await for fetchCalls to finish before fetchVisits

lspdrz:
I THINK this will fix the graphql formatError that's currently spamming sentry. At least it got rid of the error in my network tab locally, but it's hard to understand if it's the same error being reported by sentry


src/pages/Dashboard/index.tsx
```diff
@@ -16,10 +16,13 @@ const DashboardPage: React.FC = () => {
   const contacts = useAppSelector(selectAllContacts);
   const meetings = useUpcomingMeetings();
 
-  useEffect(() => {
-    dispatch(fetchCalls());
+  const fetchCallsAndVisits = async () => {
+    await dispatch(fetchCalls());
```
gabrielsaruhashi, line 20: not sure I understand why this would fix it?

gabrielsaruhashi (approved):

src/pages/Dashboard/index.tsx
lspdrz: I'm assuming that it's some sort of race condition on apollo client

PR #139, opened 2021-12-21
'Messages' header on inbox

wittejm:
Only actual change is to add the `<Layout.Header>` and wrap with an additional `<Layout>` ; the rest is spacing.
messages_page_header


gabrielsaruhashi (approved):
👍

gabrielsaruhashi:
@wittejm is there a ticket yet for the empty state design when the user selected a connection, but that inbox has no messages at all yet?

wittejm:
There is not, but that's actually what I'm working on right now. I should make a ticket...

PR #140, opened 2021-12-21
No Messages view

wittejm:
* And add a 1-pixel border on the connection header (first time this border is noticeable given the almost-white background of the no-messages component).
no_messages_view


src/components/Inbox/InboxHeader.tsx
```diff
@@ -15,18 +15,18 @@ const InboxHeader = ({ connection }: Props) => {
...
+        <div className="flex items-center gap-2">
+          <VisitorAvatar visitor={connection.visitor} size="large" />
+          <span className="font-medium text-lg">
+            {getFullName(connection.visitor)}
+          </span>
+        </div>
+        <InfoCircleOutlined
+          className="cursor-pointer"
+          onClick={() => { setViewInfo(true) }}
+          style={{ fontSize: '25px', color: '#8c8c8c' }}
```
gabrielsaruhashi, line 28: you can do className
text-lg

gabrielsaruhashi: sounds like a good use case to update tailwind or just keep using the tailwind defined styles

This one is probably bg-gray-100 or bg-gray-50?

flagging this as an important conversation topic to have with @eliz214 
so we avoid color creep

gabrielsaruhashi (approved):

PR #141, opened 2021-12-22
message cards sent by Inmate are blue; cards are rounded

wittejm:
Uses blue shades that approximate figmas; need to part of colors sync.

message_colors


gabrielsaruhashi (approved):
nice :)

PR #144, opened 2022-01-04
Message card colors and 'New' status text

wittejm:
message_color_tweak
n


tailwind.config.js
```diff
@@ -19,10 +19,17 @@ module.exports = {
       },
       red: {
         ...defaultTheme.colors.red,
-        100: '#F66262',
-        200: '#B13E37'
-      }
+        100: "#F66262",
+        200: "#B13E37"
+      },
+      primary: {
```
wittejm, line 25: These are the color names used in Figma. We're due for a sync meeting about our color palette though, and the other numbers in this sequence should probably go in here then. For now I just wanted to match the design.

cainlevy (approved):

PR #146, opened 2022-01-06
tailwind overrides antd

gabrielsaruhashi:
We want the custom tailwind styling to take precedence over antd defaults


tailwind.config.js
```diff
@@ -20,19 +20,20 @@ module.exports = {
       red: {
         ...defaultTheme.colors.red,
         100: "#F66262",
-        200: "#B13E37"
+        200: "#B13E37",
       },
       primary: {
         2: "#BAE7FF",
-        6: "#1890FF"
+        6: "#1890FF",
```
gabrielsaruhashi, line 27: @eliz214 could we adjust this naming system to follow the X00 convention?

wittejm (approved):

PR #149, opened 2022-01-06
misc cleanup: query organization, dead code, and simpler page folder

cainlevy:
The commits tell the story.


wittejm (approved):
Let's put these mutations in the mutations folder instead of queries?

cainlevy:
oh, good catch!

PR #152, opened 2022-01-07
move "contacts" loading to useQuery

cainlevy:
The Dashboard loads all connections for display. Nothing else needs them now that the meetings slice is fully hydrated.


src/pages/Dashboard.tsx
```diff
@@ -29,10 +32,16 @@ const DashboardPage: React.FC = () => {
     return () => clearInterval(interval);
   }, [dispatch]);
 
+  const { data } = useQuery(GetContactsDocument, {
+    onError: () =>
+      showToast("fetchContacts", i18n.t("api.fetchContacts"), "error"),
```
cainlevy, line 37: maintaining the previous functionality where we show a toast message on error instead of throwing

wittejm (approved):
Nice. `contact` is still used in variables name/components which should be renamed to `visitor` for naming consistency with our data model. Maybe that's worth a follow-up ticket/PR?

cainlevy:
yeah that'd be nice! i'll start on that next.

PR #156, opened 2022-01-07
useQuery for call and feedback pages

cainlevy:
Refactor `src/pages/Call.tsx` and `src/pages/CallFeedback.tsx` with `useQuery`.


src/graphql/fragments/MeetingFragment.graphql
```diff
@@ -0,0 +1,12 @@
+fragment MeetingFragment on Meeting {
+  id
+  meetingType
+  status
+  kiosk {
```
wittejm, line 5: I kind of want to get dogmatic about excluding child objects from fragments. Alternatively, is there a clear intuition on why `kiosk` and `interval` are included here but `call` is not?

cainlevy: it would be easier to be consistent about it, that's for sure.

- i included `interval` because it's a representation of primitive fields.
- i excluded `call` because i want to minimize queries against kubernetes
- i left `kiosk` because ... i didn't have a good reason

wittejm (approved):
Cool, I like it.

PR #159, opened 2022-01-07
load meetings for dashboard with useQuery

cainlevy:
I also decided to remove the profile page's usage metrics, as per https://ameelioorg.slack.com/archives/C01TGJH69V4/p1641597065024500.

After this, Redux is only used for session and UI state (full screen, modals, and routes)! 😅


src/components/Dashboard/Dashboard.tsx
```diff
@@ -35,7 +35,7 @@ const Dashboard: React.FC<Props> = ({
         <Row gutter={16}>
           <Col span={16}>
             <MeetingsList
-              meetings={meetings.slice(0, 3)}
+              meetings={meetings}
```
cainlevy, line 38: moved to the top-level data filter/sort/slice

src/pages/Call.tsx
```diff
@@ -45,19 +44,8 @@ const CallPage: React.FC<RouteComponentProps<TParams>> = ({ match }) => {
   }, [dispatch]);
 
   const leaveCall = useCallback(
-    (exitType: CallExitType) => {
-      // update the dashboard so users don't think they can come back
-      // TODO: fix by reloading data on dashboard
```
cainlevy, line 50: happening now!

wittejm (approved):
:tada:

PR #165, opened 2022-01-18
automatically opens first connection in messages page

lspdrz:
[sc-2311]


gabrielsaruhashi (approved):

src/pages/InboxPage.tsx
```diff
@@ -60,6 +60,10 @@ const InboxPage: React.FC<Props> = () => {
 
   const connections = connectionData?.currentInmate?.connections || [];
 
+  useEffect(() => {
+    if (connections) setSelected(connections[0]);
+  }, [connections]);
```
wittejm, line 65: I'd expect this to re-select the first connection every time the connections data changes, which includes message data. So when any new message sent is sent it would select the first connection again. If that is not happening, then color me confused!

lspdrz: @wittejm I think we only fetch the connections every time the user navigates to the inbox page? So it shold be ok?

gabrielsaruhashi: @wittejm we're not loading messages on this query!

wittejm: @gabrielsaruhashi @lspdrz Even though they're separate queries they can touch the same connection object in the apollo cache which is specified by the name `Connection:[objectId]`, I think. But at this point I'm speculating a bit and if it works on your end then that sounds fine.

wittejm (approved):

src/pages/InboxPage.tsx
lspdrz: @wittejm good catch!! That was indeed happening, I didn't notice that the connections were being retrieved from the messages query. I removed the dependency array element altogether, we only want this run on mount anyways

PR #166, opened 2022-01-18
Lp polling

lspdrz:
[sc-2314]


cainlevy (approved):

src/pages/InboxPage.tsx
```diff
@@ -54,10 +54,13 @@ const InboxPage: React.FC<Props> = () => {
     {
       variables: { id: selected?.id || "" },
       skip: !selected?.id,
-      nextFetchPolicy: "cache-only",
+      pollInterval: 30000,
```
cainlevy, line 57: this will refetch any currently selected connection but will not refresh the connections list. intentional?

lspdrz: that's a good catch. i was experimenting removing that while debugging something. adding back now

PR #168, opened 2022-01-19
added vertical padding on the messaging restriction banner

lspdrz:
[sc-2339]

image

@gabrielsaruhashi is this enough padding?


src/pages/InboxPage.tsx
```diff
@@ -205,7 +205,7 @@ const InboxPage: React.FC<Props> = () => {
             )}
             <Layout.Footer>
               {eMessagingRestriction ? (
-                <div className="flex flex-col items-center bg-gray-200 rounded-md">
+                <div className="flex flex-col items-center bg-gray-200 rounded-md py-1">
```
gabrielsaruhashi, line 208: could we do py-2 as recommended on the issue

PR #169, opened 2022-01-19
changed message list to hifi design

lspdrz:
[sc-2015]

DO NOT MERGE UNTIL https://github.com/Ameelio/connect-api-nest/pull/637 is merged and deployed

image


src/graphql/fragments/MessageFragment.graphql
```diff
@@ -8,9 +8,13 @@ fragment MessageFragment on Message {
   sender {
     ... on Inmate {
       id
+      firstName
+      lastName
```
gabrielsaruhashi, line 12: i dont think we're using last name, are we?
seems like only first name is used

gabrielsaruhashi (approved):

lspdrz:
@gabrielsaruhashi good catches, changed on both

PR #171, opened 2022-01-20
only use active connections on inbox page

lspdrz:
[sc-2363]


wittejm:
The figma for inbox has connections sorted into "active" and "disabled", is that no longer what we're doing here?

src/pages/InboxPage.tsx
```diff
@@ -62,7 +62,10 @@ const InboxPage: React.FC<Props> = () => {
   if (messagesLoading)
     showToast("inboxPage/loading", "Fetching the latest updates...", "loading");
 
-  const connections = connectionData?.currentInmate?.connections || [];
+  const connections =
```
gabrielsaruhashi, line 65: wrap in useMemo to avoid rerenders?

gabrielsaruhashi (approved):

gabrielsaruhashi:
@lspdrz i fixed this on a separate PR bc I needed it to unblock my work.
Gonna close this PR

PR #172, opened 2022-01-23
implement timeline on message inbox

gabrielsaruhashi:
Long PR working on the transition from messages to timeline!

[sc-2172]

Noteworthy changes:
- Deprecated MessageList
- Depracted GetConnectionMessages in favor of GetConnectionTimeline
- Updated graphql type with new definitions (kiosk and scheduled end are now optional values on the Meetings entity)

https://user-images.githubusercontent.com/12942772/150699930-29e48d53-941c-4c83-bbf7-91b06929b599.mov


src/pages/InboxPage.tsx
```diff
@@ -74,36 +106,59 @@ const InboxPage: React.FC<Props> = () => {
...
+        if (!response) return;
+        const { message, event } = response;
         cache.modify({
           id: cache.identify({
             __typename: "Connection",
-            id: newMessage.connection.id,
+            id: message.connection.id,
           }),
           fields: {
             messages(existingRef) {
```
gabrielsaruhashi, line 119: we still need the messages field to track unread messages through the MessageConnection

src/components/Modals/VoiceCallModal.tsx
```diff
@@ -26,11 +33,22 @@ const VoiceCallModal = ({ data, onCloseModal }: Props) => {
...
   const { authInfo } = useAppSelector((state) => state.session);
   const [playJoinCall] = useSound(JoinCallSound);
   const [playLeaveCall] = useSound(LeaveCallSound);
 
+  // get latest information about the video handler
+  const { data: callData, loading } = useQuery(GetCallDocument, {
+    variables: { meetingId: callId },
+    fetchPolicy: "no-cache", // do not cache access token
+    onError: (err) => reportApolloError(err, "VoiceCallModal/GetCallDocument"),
+  });
```
cainlevy, line 46: with video calls this query exists on a page component. is that a pattern we try to follow?

src/graphql/fragments/InmateFragment.graphql
```diff
@@ -0,0 +1,3 @@
+fragment InmateFragment on Inmate {
```
cainlevy, line 1: is this used?

src/pages/Call.tsx
```diff
@@ -52,11 +52,7 @@ const CallPage: React.FC<RouteComponentProps<TParams>> = ({ match }) => {
   if (error) throw new Error(error.message);
   if (loading || !data) return <Loader fullPage />;
 
-  if (
-    !data.meeting.call ||
-    data.meeting.call.status !== CallStatus.Active ||
-    new Date(data.meeting.interval.endAt) < new Date()
```
cainlevy, line 58: this previously existed as a failsafe in case the server didn't shut down a video handler properly, right?

i think we can be pretty confident in `meeting.status` these days, since a cron job checks every 5 minutes to clean those up. however i do sometimes still see video handlers fail to shut down because of network issues during the `callEnded` webhook.

is it safe to get rid of this logic, here and elsewhere?

src/pages/CallFeedback.tsx
```diff
@@ -87,23 +87,6 @@ const CallFeedbackPage: React.FC<Props> = ({ match }) => {
       />
     );
 
-  if (addMinutes(new Date(call.interval.endAt), 15) < new Date())
```
cainlevy, line 90: why do we no longer need this?

src/pages/Dashboard.tsx
```diff
@@ -31,9 +31,7 @@ const DashboardPage: React.FC = () => {
   )
     .filter(
       (m) =>
-        (m.status === MeetingStatus.Live ||
-          m.status === MeetingStatus.Scheduled) &&
-        new Date(m.interval.endAt) >= new Date()
```
cainlevy, line 36: do we still need this defensive programming?

src/pages/InboxPage.tsx
```diff
@@ -49,77 +48,100 @@ const InboxPage: React.FC<Props> = () => {
...
-      id: client.cache.identify({
-        __typename: "Connection",
-        id: selected?.id,
-      }),
-      fields: {
-        messages(existingRef) {
-          return { ...existingRef, unreadMessageCount: 0 };
-        },
-      },
-    });
```
cainlevy, line 121: where did this logic go? do we still need it?

cainlevy (commented):
this should have been multiple pull requests

src/graphql/fragments/InmateFragment.graphql
gabrielsaruhashi: it is currently only used to defined the sender props

src/pages/CallFeedback.tsx
gabrielsaruhashi: this was a check to make sure ppl are not rating a video that's been inactive for a while. 
i can add it back but I'd just make it check for start at

src/pages/InboxPage.tsx
gabrielsaruhashi: Not really! We use the timeline events to populate the inbox. unreadMessageCount is pulled from inmates directly
cc: @wittejm based on the implementation, it looks like the unreadMessageCount defined on MessageConnection is not really used based on what I saw?

src/components/Modals/VoiceCallModal.tsx
gabrielsaruhashi: This will probably turn into a page component very soon based on the early results I am seeing on Liz's work.

Re: using queries on Modals, I had originally been a strong proponent of only accessing states on higher-level containers, but the pattern got lost in previous PRs - now we do queries even on child components apparently. For practical purposes, it does seem to make sense, so I'm personally fine with it

src/components/Inbox/EventTimelineList.tsx
```diff
@@ -0,0 +1,61 @@
+import { Layout } from "antd";
+import { useEffect, useRef } from "react";
+import { GetConnectionTimelineQuery } from "src/graphql/queries/GetConnectionTimeline.generated";
+import { getMessageSenderType } from "src/utils/Message";
+import MeetingCard from "./MeetingCard";
+import MessageCard from "./MessageCard";
+import { TMessageItem } from "./MessageCard/MessageCard";
+
+type TimelineEvent =
+  GetConnectionTimelineQuery["connection"]["timelineEvents"]["edges"][0]["node"];
```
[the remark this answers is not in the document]
gabrielsaruhashi: I personally thought so as this is a one-off that's only used here?

src/utils/error.ts
```diff
@@ -0,0 +1,14 @@
+import { ApolloError } from "@apollo/client";
+import * as Sentry from "@sentry/react";
+import { showToast } from ".";
+
+export const reportApolloError = (error: ApolloError, key: string) => {
+  if (process.env.NODE_ENV === "development")
+    alert(`Dev mode alert: ${error.message}`);
```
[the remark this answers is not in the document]
gabrielsaruhashi: when we use onError, the apollo error is not bubbling up to the error boundary. 
i personally thought this accelerated dev a lot instead of going to the console and looking for the network request

src/graphql/fragments/InmateFragment.graphql
gabrielsaruhashi: alternatively I could hardcode the typename of the sender?
but this felt easier to maintain

src/pages/Call.tsx
gabrielsaruhashi: will add them back! i had assumed these checks were overkill.

src/graphql/fragments/InmateFragment.graphql
gabrielsaruhashi: oh have a new idea

src/utils/error.ts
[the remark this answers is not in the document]
cainlevy: ahh, so we're not getting the full screen stack trace when we handle the error this way. got it.

src/utils/Call.ts
```diff
@@ -49,3 +52,9 @@ export function getFullNames(
 ) {
   return visitors.map((v) => `${v.firstName} ${v.lastName}`).join(", ");
 }
+
+export function hasMeetingExpired(meeting: { interval: Interval }) {
```
cainlevy, line 56: 👍

cainlevy (approved):

PR #176, opened 2022-01-25
A first complete frontend test case

wittejm:
Paired with Lance. 

 * Provides mock apollo query data
 * Checks page state before and after the apollo query finishes loading
 * Checks modal visible/invisible states

A bit more cleanup: (got squashed into the same commit during a minor git snafu.) 

 * rename a component to match its filename
 * and some imports cleanup on InboxPage

Question for @cainlevy : In the second commit I added alternate checks for the page loading state. I feel like this could be a race condition against the apollo (mock) query resolving. If not, why are these `expect`s guaranteed to run before loading finishes?


src/graphql/queries/GetConnections.mock.ts
```diff
@@ -0,0 +1,101 @@
+// TODO:
+// Generate mock data in a generalizable and terse way
+// TODO: (related?)
+// Add this graphql type signature and resolve issues that appear:  { data: GetConnectionsQuery }
```
cainlevy, line 4: i looked through graphql code generator's typescript plugin and didn't see an option for expressing enums with the union of strings pattern.

src/graphql/queries/GetConnections.mock.ts
```diff
@@ -0,0 +1,101 @@
...
+          ],
+          __typename: "Connection",
+          messages: {
+            unreadMessageCount: 0,
+            __typename: "MessageConnection",
+          },
+          visitor: {
+            id: "5ea75f6c-2cd6-4ab9-84fb-3a7730f4cb57",
+            firstName: "Test-3",
+            lastName: "Visitor-3",
```
cainlevy, line 30: do you prefer minimalist or interesting mock data?

the minimal version might only have one connection.

src/pages/InboxPage.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+    expect(
+      screen.queryByText("Welcome to Ameelio Connect messaging!")
+    ).not.toBeVisible();
+
+    //A splash screen appears if the inmate has no connections
+    expect(screen.queryByText("No connections yet")).toBeInTheDocument();
+  });
+
+  it("populates connections and messages", () => {
+    //TODO: complete this test.
```
cainlevy, line 58: should this be included in the PR?

src/test/matchMedia.mock.ts
```diff
@@ -0,0 +1,21 @@
+// This is like doing window.matchMedia = ()=>jestMock but it lets us overwrite an unwritable field of window .
+// TODO: we'll need this for anything with a modal. Make it part of jest setup for every test.
+Object.defineProperty(window, "matchMedia", {
+  writable: true,
+  value: (query: any) => {
+    //jest.fn().mockImplementation(
+    // We tried using that, but there was an error, something like due to resetting the mock
```
cainlevy, line 7: ✂️

cainlevy (approved):
@lspdrz @gabrielsaruhashi @tanyabrassie there's a fair bit packed into this one small test case. i recommend looking at:

- how to satisfy react context in tests
- why we had to use `toBeVisible`
- how [testing-library](https://testing-library.com/) promotes writing user-focused tests

there's certainly more we can do to invest in this testing setup. let's watch how our patterns develop before we create abstractions.

src/pages/InboxPage.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    // A modal appears when loading completes if and only if the inmate has no connections
+    await waitFor(() =>
+      expect(
+        screen.getByText("Welcome to Ameelio Connect messaging!")
+      ).toBeVisible()
+    );
+
+    // The user dismisses the modal
+    fireEvent.click(screen.getByText("Get Started"));
```
tanyabrassie, line 48: I think simulating user events with the `userEvent` companion library is preferred over fireEvent since it simulates more ~ realistically ~ : https://testing-library.com/docs/guide-events/

https://testing-library.com/docs/ecosystem-user-event

src/graphql/queries/GetConnections.mock.ts
wittejm: I'll pull this mock data out to another PR separate for the second user flow. But I would lean toward interesting.... actually I would lean toward both: a data mock with one connection and another with two. But given the choice I would say two because it implies more complex behavior.

src/pages/InboxPage.spec.tsx
wittejm: Good question: yes, will remove!

src/pages/InboxPage.spec.tsx
cainlevy: cool!

PR #180, opened 2022-01-27
run tests on pull requests to main

cainlevy:
We have a test, let's run it!

Next, I will configure the repository to require that the status check passes.


wittejm (approved):
Woohoo!

PR #182, opened 2022-01-31
call confirm popover with times remaining

wittejm:
Requires this api change: https://github.com/Ameelio/connect-api-nest/pull/681

https://user-images.githubusercontent.com/2104990/151882440-b9be063b-546a-49f6-85d4-551c8908354d.mov


src/components/Inbox/VoiceCall/StartVoiceCallPopover.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const StartVoiceCallPopover: React.FC<Props> = ({
+  onConfirm,
+  onClose,
+  isOpen,
+  children,
+}) => {
+  const { data } = useQuery(GetVoiceCallTimesDocument, {
+    onError: (err) =>
+      reportApolloError(err, "StartVoiceCallPopover/GetVoiceCallTimesDocument"),
```
cainlevy, line 20: slight improvement from earlier today:

```suggestion
    onError: errorReporter("StartVoiceCallPopover/GetVoiceCallTimesDocument"),
```

src/components/Inbox/VoiceCall/StartVoiceCallPopover.tsx
```diff
@@ -0,0 +1,59 @@
...
+  });
```
cainlevy, line 21: should we customize the network policy for this query to manage caching?

on the one hand, it seems like time sensitive data that could get stale quickly. on the other hand, we don't (currently) expect pathways clients to remain logged in for long.

src/utils/Time.ts
```diff
@@ -17,3 +17,13 @@ export function intervalSecondsToMinutesLabel(seconds?: number | null): string {
...
 }
+
+export function millisecondsToMinutesAndSeconds(
+  milliseconds: number | undefined
+): string {
+  if (!milliseconds) return "";
+  const minutes = Math.floor(milliseconds / (60 * 1000));
+  const remainder = milliseconds - minutes * 60 * 1000;
+  const seconds = `${Math.floor(remainder / 1000)}`.padStart(2, "0");
+  return `${minutes}:${seconds}`;
```
cainlevy, line 28: non-blocking UX thought: `seconds` may be a distracting and unhelpful amount of granularity

cainlevy (approved):

src/components/Inbox/VoiceCall/StartVoiceCallPopover.tsx
wittejm: I considered using `refetch` somewhere but didn't see a clever place to put it.

Then I started writing this comment above and thought of a clever place to put it, so, thanks!

src/utils/Time.ts
wittejm: Makes sense! The ticket is explicitly calling it low-fi so I'll leave this as is.

src/components/Inbox/VoiceCall/StartVoiceCallPopover.tsx
wittejm: I see that's in DOC client, not Pathways yet.

cainlevy: yep my bad!

PR #183, opened 2022-02-03
connect-call-client v0.3.0

cainlevy:
v0.3.0 of connect-call-client requires the application to decide when/which tracks it wants to produce.


cainlevy:
verified locally that pathways can still join calls

wittejm (approved):

PR #186, opened 2022-02-07
permit login form with AK

cainlevy:
Navigating to `https://pathways.ameelio.org/#/login/AK` would pre-populate the state field, but not in a way that passes validations.

We don't really need to validate the state, however. We only use it to narrow down the list of facilities.


wittejm (approved):
...AntD doesn't validate based on the current value of `value`? Hrmm.

cainlevy:
Yeah I think there's some disconnect there when the value isn't represented by the options.

PR #187, opened 2022-02-07
fix date scalar type, for all outputs

cainlevy:
1. Date scalar config should apply to all generated outputs. This is an indentation change.
2. Fix the type to be `number`. The `string` type caught some bug but was still wrong.


wittejm (approved):
Really nice that `Date()` just handles this switch without issue.

PR #188, opened 2022-02-07
fix button transition to "Join Now"

cainlevy:
Previously the "Join Now" button would only appear if the last component render happened after the meeting start date. However the meeting status goes "live" early, and the call handler goes "active" early, so the only thing left to do is wait for the clock to tick forward a bit.

This is entirely dependent on synchronized clocks! Pathways clients with clock drift can still experience odd results.


src/components/Dashboard/MeetingItem.tsx
```diff
@@ -1,51 +1,60 @@
...
-        return "#F66262";
-      case MeetingType.InPersonNoContact:
-        return "#B13E37";
-    }
-  };
+  useEffect(() => {
+    if (afterStart) return;
+    const timer = setTimeout(
+      () => setAfterStart(true),
+      meeting.interval.startAt - Date.now()
```
cainlevy, line 44: oh and as always, javascript will integer overflow for values greater than about 24.8 days. are we safe?

https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value

gabrielsaruhashi (approved):

PR #194, opened 2022-02-15
imports json directly on test file

gabrielsaruhashi:


cainlevy (approved):
cool! so this test will additionally verify that i18n works and not break on copy changes.

PR #195, opened 2022-02-15
Fetch last events for connection timeline

wittejm:
I'm assuming we want to fetch 500. 

I tested correctness by trying this with the argument `first: 5` and with `last: 5`


src/graphql/queries/GetConnectionTimeline.graphql
```diff
@@ -5,7 +5,7 @@ query GetConnectionTimeline($id: ID!) {
     visitor {
       ...VisitorFragment
     }
-    timelineEvents {
+    timelineEvents(last: 500) {
```
cainlevy, line 8: 500 seems fine with me! that should delay our need to build "load older" functionality. once we have "load older", we can tune this to a smaller page size.

cainlevy (approved):

PR #196, opened 2022-02-24
Start Voice Call modal 

wittejm:
Requires: https://github.com/Ameelio/connect-api-nest/pull/742

https://user-images.githubusercontent.com/2104990/155597529-15019bb9-faff-4c2c-a9aa-555d99811914.mov


src/components/Inbox/VoiceCall/StartVoiceCallModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const [isPanelOpen, setIsPanelOpen] = useState(0);
+
+  const privacyNotice = () => (
+    <div>
+      <div>
+        <span className="text-xs text-black opacity-45">PRIVACY NOTICE</span>
+      </div>
+      <div className="p-3 m-2 border border-gray-100">
+        By participating in this call, I consent to the monitoring and recording
```
wittejm, line 35: woops didn't i18n this.

src/components/Inbox/VoiceCall/StartVoiceCallModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const remainingDuration = millisecondsToMinutes(
+    data?.currentInmate?.remainingVoiceCallDuration || 0
+  );
+  const weeklyRemainingQuota = millisecondsToMinutes(
+    data?.currentInmate?.voiceCallTimeQuota || 0
+  );
+  const maxDuration = Math.floor(
+    (data?.currentInmate?.facility.maxVoiceCallDuration || 0) / (60 * 1000)
+  );
+  const resetMinutes = 60 - maxDuration;
```
cainlevy, line 50: this `60` is a shared assumption with the backend, right?

https://github.com/Ameelio/connect-api-nest/blob/458119495c5c51188ccbc969a938e866915de00c/src/meetings/getRemainingVoiceCallDuration.ts#L16

src/components/Inbox/VoiceCall/StartVoiceCallModal.tsx
```diff
@@ -0,0 +1,206 @@
...
+                      <div className="italic">
+                        {t("startVoiceCallModal.sessionLimitReset", {
+                          resetMinutes,
+                        })}
+                      </div>
+                    </div>
+                  ),
+                },
+                {
+                  // TODO: This copy is not general to weekly quota amounts other than 2 hours.
```
cainlevy, line 160: out of scope for this pull request or for the voice calls mvp?

src/i18n/es/modals.json
```diff
@@ -30,5 +30,28 @@
     "monitoring": "Todas las llamadas son monitoreadas y grabadas por la prisión por razones de seguridad.",
     "calling": "Llamando",
     "mute": "Silenciar"
+  },
+  "startVoiceCallModal": {
+    "title": "",
```
cainlevy, line 35: cc @gabrielsaruhashi because i can't remember if you have a preference between empty or english defaults

src/utils/Time.ts
```diff
@@ -27,3 +27,11 @@ export function millisecondsToMinutesAndSeconds(
   const seconds = `${Math.floor(remainder / 1000)}`.padStart(2, "0");
   return `${minutes}:${seconds}`;
 }
+
+export function millisecondsToMinutes(
```
cainlevy, line 31: i'm surprised date-fns didn't have something like this

cainlevy (approved):

src/components/Inbox/VoiceCall/StartVoiceCallModal.tsx
wittejm: Ah yea, I can name a constant here too to make it a little more clear.

src/components/Inbox/VoiceCall/StartVoiceCallModal.tsx
wittejm: A good question for @gabrielsaruhashi !

src/utils/Time.ts
wittejm: LTM like date-fns does stuff with dates but doesn't really do stuff with time units.

src/i18n/es/modals.json
gabrielsaruhashi: English would be good here (just in case we forget to translate). It will be better than showing empty strings

wittejm: Gotcha!

src/components/Inbox/VoiceCall/StartVoiceCallModal.tsx
gabrielsaruhashi: @wittejm this should be dynamic and based on the inc person's quota!

wittejm: Just to be sure: you mean their maximum quota? This copy shouldn't change as the inmate consumes their own quota, correct?

wittejm:
Generalized to any weekly quota amount and it reports as "x hours (y minutes)" only if it's a round hour amount.

I looked at date-fn's `formatDuration` and it doesn't actually do what I was looking for. I wrote the i18n model to handle the "hours" vs "minutes" cases. 

I simplified the util function to return a number instead of a string, and pulled it into a local definition because it's so simple and probably not good for general use, and differs from the other util function which builds a string. 

ready for re-review @gabrielsaruhashi @cainlevy

src/components/Inbox/VoiceCall/StartVoiceCallModal.tsx
```diff
@@ -38,16 +38,42 @@ const StartVoiceCallModal: React.FC<Props> = ({
...
+  const resetMinutes = MAX_VOICE_CALL_DURATION_PERIOD_MINUTES - maxDuration;
+
+  const weeklyQuotaRuleHeader =
+    maxTimeQuota % 60 === 0
+      ? t("startVoiceCallModal.weeklyQuotaRuleHeaderHours", {
+          hours: maxTimeQuota / 60,
+        })
+      : t("startVoiceCallModal.weeklyQuotaRuleHeaderMinutes", {
+          minutes: maxTimeQuota,
+        });
```
cainlevy, line 67: code golf idea: it's okay to pass extra keys to translation so you could reuse a `{hours, minutes}` object for all of these.

cainlevy (approved):

PR #199, opened 2022-03-01
bump nodejs to v16

cainlevy:
NodeJS v16 is the current LTS


Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN dnf install -y glibc-langpack-en \
 RUN dnf module enable -y nginx:1.18 \
  && dnf install -y \
     nginx \
- && dnf module install -y nodejs:14 \
+ && dnf module install -y nodejs:16 \
```
cainlevy, line 28: is this all we need to manage the deployed runtime?

FreedomBen: Yep as long as that module exists (or in other words, if it builds with this change), that's all it should take.  Is v16 an LTS?    I'll try this locally and see if it works

FreedomBen: @cainlevy yep it looks good :+1:

FreedomBen (approved):

PR #200, opened 2022-03-03
blocks emojis

gabrielsaruhashi:
Need to add a filter server-side as well.

https://user-images.githubusercontent.com/12942772/156624963-02d6eef2-4367-44ab-b29e-58b67b0d47b9.mov


src/components/Inbox/MessageCompose.tsx
```diff
@@ -31,7 +32,18 @@ const MessageCompose = ({
       <Input.TextArea
         value={draft}
         rows={1}
-        onChange={(e) => setDraft(e.target.value)}
+        onChange={(e) => {
+          const { value } = e.target;
+          const clean = value.replace(/[^\x00-\x7F|àèòùáéíóúñÑÁÉÍÓÚ¿¡]/g, "");
```
wittejm, line 37: neat. Does this replace those individual accented characters as well?

wittejm (approved):

src/i18n/en/messaging.json
```diff
@@ -28,6 +28,9 @@
     "messageAvailability": {
       "pending": "Review Required",
       "delivered": "Delivered"
+    },
+    "errors": {
+      "emojiNotAllowed": "Due to DOC restrictions, emojis and special characters are not allowed 😔."
```
cainlevy, line 33: haha, rubbing it in by using an emoji.

should we have neutral copy here? i think we can probably point fingers without apologizing.

src/components/Inbox/MessageCompose.tsx
cainlevy, line 37: 👍 for an allow list

cainlevy (approved):

src/components/Inbox/MessageCompose.tsx
gabrielsaruhashi: there's a [caret] so it is not a NOT operator. 
accents are pretty common in Spanish and other romance languages so I thought we should allow them
Although the downside is that blocking them would probably better for the message detection system.

src/i18n/en/messaging.json
gabrielsaruhashi: that's fair - let me check with Liz what she thinks! I'll add this note

gabrielsaruhashi: will merge it to keep the ball rolling and will revisit copy as needed once the feedback is in on slack

PR #201, opened 2022-03-04
when users click on the connection card on the main dash, they open the connection drawer

gabrielsaruhashi:
https://user-images.githubusercontent.com/12942772/156767887-f7b605fa-42c2-4aaa-af6f-b1b71052c627.mov


src/graphql/fragments/ConnectionFragment.graphql
```diff
@@ -1,6 +1,6 @@
 fragment ConnectionFragment on Connection {
   id
-  status
+  connectionStatus: status
```
gabrielsaruhashi, line 3: had to add an alias to this field because ConnectionStatus was conflicting w/ MeetingStatus on the GetConnectionTimeline query

src/graphql/queries/GetConnectionTimeline.graphql
```diff
@@ -10,6 +10,9 @@ query GetConnectionTimeline($id: ID!) {
         node {
           ...ConnectionTimelineEventFragment
           object {
+            ... on Connection {
+              ...ConnectionFragment
```
gabrielsaruhashi, line 14: seems like we hadnt updated the graphql.ts since this new timeline event was added

src/graphql/queries/GetVisitors.graphql
```diff
@@ -2,11 +2,13 @@ query GetVisitors {
   currentInmate {
     id
     connections {
-      id
-      status
+      ...ConnectionFragment
       visitor {
         ...VisitorFragment
       }
+      activeRestrictions {
```
gabrielsaruhashi, line 9: info needed for the connection drawer

src/components/Dashboard/ConnectionsList.tsx
```diff
@@ -28,9 +23,12 @@ const ConnectionsList: React.FC<Props> = ({ connections }: Props) => {
...
   ];
 
-  const visitors = connections
-    .filter((c) => c.status === activeTab)
-    .map((c) => c.visitor);
+  const numConnections = connections.length;
+
+  const filteredConnections = useMemo(
+    () => connections.filter((c) => c.connectionStatus === activeTab),
+    [activeTab, numConnections]
```
cainlevy, line 30: depending on `connections` may be slightly better because it should allow the memo to update if connection details change, like name or status.

cainlevy (approved):

PR #204, opened 2022-03-04
can toggle time alerts on and off

gabrielsaruhashi:
[sc-2903]

image
image


src/components/Inbox/VoiceCall/VoiceCallInterface.tsx
```diff
@@ -147,10 +149,33 @@ const VoiceCallInterface = ({ vc, authInfo, onTerminate }: Props) => {
...
+          <IconButton
+            icon={<ClockCircleOutlined className="text-lg" />}
+            type={hasTimeAlert ? "primary" : "default"}
+            ghost={hasTimeAlert}
+            shape="circle"
+            onClick={() => setHasTimeAlert((timeAlert) => !timeAlert)}
+            label={`${t("call:videoOverlay.timeAlert.label")}`}
+            className={
+              hasTimeAlert
+                ? "text-blue-400 border-blue-400"
```
gabrielsaruhashi, line 173: this is needed to override the default "press" state color that antd offers

src/components/Inbox/VoiceCall/VoiceCallInterface.tsx
```diff
@@ -147,10 +149,33 @@ const VoiceCallInterface = ({ vc, authInfo, onTerminate }: Props) => {
...
+              : t("call:videoOverlay.timeAlert.tooltipAlertOff")
+          }
+        >
+          <IconButton
+            icon={<ClockCircleOutlined className="text-lg" />}
+            type={hasTimeAlert ? "primary" : "default"}
+            ghost={hasTimeAlert}
+            shape="circle"
+            onClick={() => setHasTimeAlert((timeAlert) => !timeAlert)}
+            label={`${t("call:videoOverlay.timeAlert.label")}`}
```
cainlevy, line 170: 🏌️‍♂️ 

```suggestion
            label={t("call:videoOverlay.timeAlert.label")}
```

src/components/Inbox/VoiceCall/VoiceCallInterface.tsx
cainlevy: should we bake this color choice into the common component? or are there other places in the app that aren't ready for an update.

cainlevy (approved):

src/components/Inbox/VoiceCall/VoiceCallInterface.tsx
gabrielsaruhashi: this behavior is pretty specific to this implementation bc this button is acting like a toggle :/ 
but yes, we should move to the common component now that liz and jordan are wrapping up the color updates

PR #205, opened 2022-03-07
should prompt the pathways user to rate the call

gabrielsaruhashi:
image
image


src/components/CallFeedback/VideoCallRatingCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+          </Button>
+        ) : (
+          <Button
+            size="large"
+            disabled={hasIncompleteRating}
+            onClick={() => onSubmit({ rating, feedback })}
+          >
+            {t("common:logout")}
+          </Button>
+        )}
```
wittejm, line 55: looks like the only difference in this ternary is the text.

src/components/Inbox/VoiceCall/VoiceCallInterface.tsx
```diff
@@ -106,11 +108,48 @@ const VoiceCallInterface = ({ vc, authInfo, onTerminate }: Props) => {
...
+          rateCall({
+            variables: {
+              input: { meetingId: vc.id, rating, description: feedback },
+            },
+          });
+          onFinish();
+          showToast(
+            "VoiceCallInterface/onSubmit",
+            t("feedback:qualityRating.submissionFeedback"),
+            "success"
```
wittejm, line 146: translate this too?

wittejm (approved):

src/components/Inbox/VoiceCall/VoiceCallInterface.tsx
gabrielsaruhashi: oh this one is just the toast type - no need to translate

PR #206, opened 2022-03-08
Hide and feature flag inactive restrictions

wittejm:
requires https://github.com/Ameelio/connect-api-nest/pull/779.

Having these two types of FacilityFeature and FacilityService is :frowning_face:


src/components/Inbox/ConnectionInfo.tsx
```diff
@@ -45,32 +43,29 @@ const ConnectionInfo = ({ connection, open, onClose }: Props) => {
...
+          {Object.keys(services).map(
+            (service) =>
+              ((service === FacilityService.VoiceCall &&
+                useFacilityFeature(FacilityFeature.VoiceCalls)) ||
+                (service === FacilityService.Call &&
+                  useFacilityFeature(FacilityFeature.VideoCalls)) ||
+                (service === FacilityService.Emessaging &&
+                  useFacilityFeature(FacilityFeature.Emessaging)) ||
+                service === FacilityService.InPersonContact ||
+                service === FacilityService.InPersonNoContact) && (
```
cainlevy, line 55: yeahhhh i see what you mean.

one thing that could make this more readable right now is extracting the filtered list of services. i found it a little challenging to find where the conditional ended.

another idea might be a `useFacilityService()` hook that extracts the filter condition itself?

i don't know what it'd take to resolve the underlying mismatches between services and features, but i imagine it's out of scope for everyone right now.

cainlevy (approved):

PR #212, opened 2022-03-11
fix hooks when facility hasMessaging

cainlevy:
React gets uppity with variable numbers of hooks. We have to run each of them every time.

Normally this is caught with ESLint, but somewhere along the way we appear to have lost the react hook linting rules. I will follow up on that separately.


cainlevy:
Fixes https://sentry.io/organizations/ameelio/issues/3090104749/?project=6021907&referrer=slack

wittejm (approved):

PR #214, opened 2022-03-11
Fix ESLint (and at least one bug)

cainlevy:
We haven't been running any of create-react-app's ESLint rules. It looks like the `.eslintrc.js` file was overwriting, instead of extending.

I've restored ESLint and fixed all errors and warnings.


src/components/Inbox/ConnectionInfo.tsx
```diff
@@ -14,27 +14,31 @@ interface Props {
...
 
-const ConnectionInfo = ({ connection, open, onClose }: Props) => {
-  const services = {
-    [FacilityService.Call]: "Video calls",
-    [FacilityService.InPersonNoContact]: "In-person visits, no contact",
-    [FacilityService.InPersonContact]: "In-person visits, with contact",
-    [FacilityService.Emessaging]: "Messaging",
-    [FacilityService.VoiceCall]: "Voice Call",
-  };
+function useFacilityServices(): FacilityService[] {
```
cainlevy, line 25: this logic is a little awkward but obeys the rules of hooks 😢

wittejm (approved):

PR #215, opened 2022-03-11
Jw/sc 2828/disable call button on restricted connection

wittejm:
Screen Shot 2022-03-11 at 3 10 19 PM

The last commit tweaks the time format used only in one other place: message timestamps. E.g. it was showing "15:00 PM". 

I don't like the name `datetimea` very much but I like it more than `datetimeampm` or `datetimeAMPM` and it's easy enough to read in the file what it does, imo.


src/graphql/queries/GetFacilityPublicId.graphql
```diff
@@ -0,0 +1,8 @@
+query GetFacilityPublicId {
+  currentInmate {
+    id
+    facility {
+      publicId
+    }
```
cainlevy, line 6: this facility should have an `id` for merging.

src/components/Inbox/InboxHeader.tsx
```diff
@@ -38,17 +58,28 @@ const InboxHeader = ({ connection, onCall, callIsLoading }: Props) => {
...
+                  ? "until " +
+                    formatDate(
+                      new Date(voiceCallRestriction.expiresAt),
+                      "datetimea"
+                    )
+                  : "indefinitely"
+              }`}
+            >
+              {
+                " " /* Oh right, these tooltips don't render without an invisible space, grumble. */
```
cainlevy, line 76: 👀

cainlevy (approved):

PR #216, opened 2022-03-14
hide all non-live, expired calls

wittejm:
This is minimal to fix the reported bug. It seems like the same data error that produced a scheduled call in the past could also produce a `LIVE` call in the past, so maybe this filter should just be `!hasMeetingExpired(m)`?


cainlevy (approved):
oh. huh. it's hard to tell what that filter was trying to do. my instinct here is to have `(live || scheduled) && !expired)`.

we do need a meeting status filter of some kind. hmm, maybe we've been relying on `InmatesResolver.meetings` to filter by status. looks like that only returns `(live || scheduled || pending)`.

okay, yeah, given the resolver's default behavior maybe it does make sense for the client to only filter on `!hasMeetingExpired`.

> It seems like the same data error that produced a scheduled call in the past could also produce a LIVE call in the past

given that the data error was manual editing, yes. however the workflow is specific to scheduling a test call and other statuses like LIVE would be compound mistakes.

PR #217, opened 2022-03-14
sc-3153 waiting for call rating to complete before logout

jackofsometrade:


src/components/CallFeedback/CallFeedback.tsx
```diff
@@ -27,7 +29,10 @@ export type CallExitType = "no_show" | "early_exit" | "ended" | "terminated";
 
 interface Props {
   call: Pick<Call, "id" | "visitors">;
-  rateCall: (rating: number, feedback?: string) => void;
+  rateCall: (
+    rating: number,
+    feedback?: string
+  ) => Promise<FetchResult<CreateCallReportMutation, Record<string, any>>>;
```
cainlevy, line 35: since this component doesn't care about the results of `rateCall()` it would be 👌 to simplify as `Promise<void>`. typescript would still allow `rateCall` to have a more specific return signature, but this component wouldn't be able to use it.

cainlevy (approved):
more ratings! 🎉

src/components/CallFeedback/CallFeedback.tsx
jackofsometrade: Sounds good, I agree

jackofsometrade: @cainlevy changed to `unknown` lemme know if that's acceptable to you

PR #219, opened 2022-03-15
connection create timestamp in timeline view

wittejm:
Screen Shot 2022-03-15 at 11 35 27 AM

https://www.figma.com/file/IXcl41KdiXJRP4QRx7TKHU/CMob%2FPath%3A-Voice-Call-Demo-%26-E-Messaging-MVP?node-id=344%3A36896


src/components/Inbox/EventTimelineList.tsx
```diff
@@ -53,7 +85,7 @@ const EventTimelineList = ({ events, recipientFirstName }: Props) => {
           // TODO: create high-fi cards for these two timeline events
           // https://app.shortcut.com/ameelio/story/3029/high-fi-cards-for-new-timeline-events
           case "ConnectionRestriction":
-          case "Connection":
+          case "Connection": // The "create" event is specially rendered at the top of the view, regardless of its order in the timeline
```
cainlevy, line 88: what might cause it to not be at the top of the timeline?

src/components/Inbox/EventTimelineList.tsx
```diff
@@ -25,9 +27,39 @@ const EventTimelineList = ({ events, recipientFirstName }: Props) => {
...
+  );
+  const createConnectionEvent = createConnectionEvents[0];
+
+  const otherEvents = events.filter(
+    (e: any) =>
+      e &&
+      e.object &&
+      (e.object.__typename !== "Connection" ||
+        e.verb !== ConnectionTimelineEventVerb.Create)
+  );
```
cainlevy, line 45: could also rely on the `switch/case` to filter this out 🏌️

cainlevy (approved):
it'd be nice if this event had a natural place in the timeline like the rest, but not a blocker

src/components/Inbox/EventTimelineList.tsx
wittejm: ooh good call

wittejm:
> it'd be nice if this event had a natural place in the timeline like the rest, but not a blocker

What do you mean? It seems like the top of the view is a natural place!

cainlevy:
> It seems like the top of the view is a natural place!

yeah, i agree. i'm was simply unsure why we need the added complexity of pulling it out from its natural place at the top of the list in order to place it at the top of the list.

it's the kind of thing that someone else in six months will assume is very intentional and fixes a problem they don't know about. that's what i assumed just now.

wittejm:
Hmm I see what you mean! I redid it, it renders the same, and I like the resulting code. I was initially caught up on "what if the data is bad" but in fact we can reasonably expect that the data will not be bad.

PR #220, opened 2022-03-15
data collection: uchicago refactoring

gabrielsaruhashi:
## Description

We used to only prompt users after the call ended. This didn't yield great results. After the call ends people tend to be walked out of the room. 

This data collection is crucial for Ameelio's success & our impact measurement. We'll be more opinionated moving forward - it's a one-time survey that will impact our ability to fundraise and publish studies on the impact of our interventions

## Caveats

This assumes that the typeform domain is whitelisted. I won't merge until we have a follow-up PR that addresses the case in which the DoC did not whitelist the domain.

We're using an embed system that I am pretty sure will cause major headaches if DoC did not whitelist this properly.

### Optional 

When the inc person has <= 3 meetings through Connect, we only surface an optional banner:

https://user-images.githubusercontent.com/12942772/158476317-856c695c-2c40-43dd-b62c-3911dce588d4.mov

### Required
When the inc person has > 3 meetings through Connect, we make it required:

https://user-images.githubusercontent.com/12942772/158478235-d079d2cf-e1c9-4c25-a789-f420ae2d8784.mov

[sc-3179]


src/components/Modals/Modals.tsx
```diff
@@ -39,19 +34,6 @@ const Modals: React.FC = () => {
           data={data}
         />
       );
-    case "KIOSK_CONFIRMATION_MODAL":
```
gabrielsaruhashi, line 42: extracting this out of redux 😵

src/components/CallFeedback/CallFeedback.tsx
```diff
@@ -76,36 +72,12 @@ const NoShowFeedback = ({
...
-  logout,
-  onSubmitIntakeForm,
-  promptIntakeForm,
-}) => {
+const CallFeedback: React.FC<Props> = ({ call, rateCall, type, logout }) => {
   const { t } = useTranslation(["feedback", "common"]);
 
   const navigate = useAppNavigate();
 
   const ExitFeedback = () => {
```
gabrielsaruhashi, line 80: we will start prompting users to fill out the survey pre-call moving fwd

src/components/Card/SurveyCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+import Card from ".";
+import SurveyButton from "../Common/Buttons/SurveyButton";
+
+type Props = {
+  title: string;
+  body: string;
+  onClose?: () => void;
+  illustration: React.ReactNode;
+  formId: string;
+};
```
cainlevy, line 13: does this selection of props imply a possible second use case for `<SurveyCard />`?

src/hooks/useIntakeForm.ts
```diff
@@ -1,21 +1,55 @@
...
-    if (user.hasFilledIntakeForm || !endedCalls) return false;
+export const useIntakeForm = (): {
+  intakeFormStatus: IntakeFormStatus;
+  updateIntakeFormStatus: () => Promise<void>;
+} => {
+  const { data: meetingsData } = useQuery(GetMeetingsDocument, {
+    onError: (e) => {
+      reportApolloError(e, "useIntakeForm/GetMeetingsDocument");
+    },
+  });
```
cainlevy, line 22: i thought this query only returned upcoming meetings. is that enough to decide if someone is a seasoned user?

i'm trying to figure out how this worked before. it looks like we have a GetFeedbackRequirements query. should that still be involved, or deleted?

cainlevy (commented):
the various files for this feature are organized by type into different locations:

- `src/components/Card/SurveyCard`
- `src/components/Common/Buttons/SurveyButton`
- `src/components/Modals/IntakeFormModal`
- `src/graphql/queries/GetIntakeFormStatus`
- `src/hooks/useIntakeForm`

i wonder how much could/should be organized together into a feature folder?

src/hooks/useIntakeForm.ts
gabrielsaruhashi: get feedback requirement is still used to load the information for the meeting ! it's also pulling ended calls but that is not being used anymore :) - good catch will update
image

gabrielsaruhashi: @cainlevy i changed that! 
created a new query "GetEndedCalls"

src/api/graphql.ts
```diff
@@ -423,6 +424,7 @@ export type Inmate = {
   profileImageUrl?: Maybe<Scalars["String"]>;
   remainingVoiceCallDuration: Scalars["Int"];
   remainingVoiceCallTimeQuota: Scalars["Int"];
+  totalMeetings: Scalars["Int"];
```
cainlevy, line 427: is this one new? i don't recognize it.

src/components/ChicagoSurvey/useIntakeForm.ts
```diff
@@ -0,0 +1,55 @@
...
+  const { data: meetingsData } = useQuery(GetEndedCallsDocument, {
+    onError: (e) => {
+      reportApolloError(e, "useIntakeForm/GetMeetingsDocument");
+    },
+  });
+  const { data: currentInmateData } = useQuery(GetIntakeFormStatusDocument, {
+    onError: (e) => {
+      reportApolloError(e, "useIntakeForm/GetIntakeFormStatusDocument");
+    },
+  });
```
cainlevy, line 27: is there a benefit to making these separate queries?

cainlevy (approved):

src/api/graphql.ts
gabrielsaruhashi: oop - dropped it

tanyabrassie:
@gabrielsaruhashi are we worried at all that if we make the survey mandatory, the data will be skewed as people might just rush through because they want to get to their call?

gabrielsaruhashi:
hi @tanyabrassie that's def a valid concern, but we agreed with UChicago that it's worth learning . the uptake right now is concerningly small given the importance of the data

PR #222, opened 2022-03-17
app recovers if typeform does not work

gabrielsaruhashi:


src/components/ChicagoSurvey/useIntakeForm.ts
```diff
@@ -43,5 +75,12 @@ export const useIntakeForm = (): {
     });
   }, [updateProfile, inmate?.id]);
 
-  return { intakeFormStatus, updateIntakeFormStatus };
+  return {
+    intakeFormStatus,
+    updateIntakeFormStatus,
+    formId:
+      intakeFormStatus !== "unavailable"
```
gabrielsaruhashi, line 82: unavailable is already checking to ensure that the env var is available

src/components/Dashboard/Dashboard.tsx
```diff
@@ -21,21 +22,20 @@ interface Props {
...
-  showSurveyBanner,
 }) => {
   const [selectedMeeting, setSelectedMeeting] = useState<Meeting | null>(null);
   const [selectedConnection, setSelectedConnection] = useState<Connection>();
   const { t } = useTranslation(["dashboard"]);
-
-  const FORM_ID = process.env.REACT_APP_INTAKE_FORM;
+  const { intakeFormStatus, formId } = useIntakeForm();
+  const showSurveyBanner =
+    intakeFormStatus === "required" || intakeFormStatus === "optional";
```
cainlevy, line 38: seems like it may be easier to return a boolean from `useIntakeForm` rather than the internal status?

src/components/ChicagoSurvey/useIntakeForm.ts
cainlevy: or maybe the presence of `formId` is really all you need to know if the survey is available and not completed

src/components/ChicagoSurvey/useIntakeForm.ts
```diff
@@ -22,12 +26,40 @@ export const useIntakeForm = (): {
...
-  const intakeFormStatus: IntakeFormStatus = useMemo(() => {
+  const isSurveyAvailable = useCallback(async () => {
+    const FORM_ID = process.env.REACT_APP_INTAKE_FORM;
+    if (!FORM_ID) return false;
+
+    const TIMEOUT_MS = 10000;
+    try {
+      await Promise.race([
+        fetch(
+          `https://form.typeform.com/to/${process.env.REACT_APP_INTAKE_FORM}`
```
cainlevy, line 37:
```suggestion
          `https://form.typeform.com/to/${FORM_ID}`
```

src/components/ChicagoSurvey/useIntakeForm.ts
```diff
@@ -43,5 +75,12 @@ export const useIntakeForm = (): {
     });
   }, [updateProfile, inmate?.id]);
 
-  return { intakeFormStatus, updateIntakeFormStatus };
+  return {
+    intakeFormStatus,
+    updateIntakeFormStatus,
```
cainlevy, line 80: 🤔 naming is a bit odd here. this doesn't update `intakeFormStatus`, but rather `inmate.hasFilledIntakeForm`. maybe it's an `onComplete` handler, or similar?

cainlevy (approved):
wow github pull requests are misbehaving today

PR #223, opened 2022-03-17
Jw/sc 3029/high fi cards for new timeline events

wittejm:
Screen Shot 2022-03-17 at 10 46 17 AM


src/components/Inbox/MeetingCard.tsx
```diff
@@ -1,29 +1,67 @@
...
+            {`${
+              meeting.meetingType === MeetingType.VoiceCall &&
+              meeting.status === MeetingStatus.NoShow
+                ? "Missed "
+                : ""
+            } ${getVisitLabel(meeting.meetingType)} •
+            ${
+              meeting.meetingType === MeetingType.VoiceCall
+                ? millisecondsToHoursOrMinutesLabel(meeting.call?.duration)
+                : !!meeting.interval.seconds
```
wittejm, line 56: admittedly, some of this logic gets a little tortured with the number of different cases. Voice call duration is the interesting value here, but for scheduled meetings it's nice to see the round numbers seen in `interval`

cainlevy: yeah it's starting to look like voice calls could have their own card

src/graphql/queries/GetConnectionTimeline.graphql
```diff
@@ -26,6 +26,11 @@ query GetConnectionTimeline($id: ID!) {
             }
             ... on Meeting {
               ...MeetingFragment
+              call {
+                ... on Call {
+                  ...CallFragment
+                }
```
cainlevy, line 32: shouldn't need the `... on Call` here.

i'm a little conflicted here. it's nice to use the fragment, but we only need duration, but i could probably add a query optimization to the backend so the other fields are less expensive ... yeah i'll just optimize the resolver.

src/utils/Time.ts
```diff
@@ -27,6 +27,16 @@ export function intervalSecondsToMinutesLabel(seconds?: number | null): string {
...
 }
 
+// Like the above, but round hour amounts are rendered as such.
+export function millisecondsToHoursOrMinutesLabel(
+  milliseconds?: number | null
+): string {
+  const mins = milliseconds && Math.round(milliseconds / (60 * 1000));
+  const hours = mins && mins % 60 === 0 ? mins / 60 : null;
+  if (hours) return `${hours} ${hours === 1 ? "hour" : "hrs"}`;
+  return `${mins} ${mins === 1 ? "min" : "mins"}`;
```
cainlevy, line 37: does this need to translate? hmm, that could be existing tech debt in our time utils.

cainlevy (approved):

src/components/Inbox/MeetingCard.tsx
wittejm: Ah yea that will help

src/graphql/queries/GetConnectionTimeline.graphql
wittejm: ^^ fragments are supposed to be cheeeap!

PR #226, opened 2022-03-18
debugging for unhandled login failures

cainlevy:
We're getting reports of Pathways login issues. I haven't found a cause so I'm trying to increase visibility.

I really didn't expect `Cannot find header token` to show up so quickly or so often. It appears to coincide with a 400 status code, but that's about all I can see.

I believe that this logging is safe and will not contain any authentication tokens. NestJS appears to only send those in response headers, not the response body.


jackofsometrade:
> I believe that this logging is safe and will not contain any authentication tokens. NestJS appears to only send those in response headers, not the response body.

I had the same concern.  If the auth header is missing, but the API still populates the user response, would that mean this error is being logged in Sentry or elsewhere, with personally identifiable info?

jackofsometrade:
I'll have to defer to your better judgement on this one.  I don't have enough experience with the API or NestJS in general to know what the response could possibly contain.

cainlevy:
It's possible that the response contains data from the inmate table. I don't understand what's happening here either.

PR #228, opened 2022-03-29
Add minutes and message to feedback flow

wittejm:
Screen Shot 2022-03-29 at 9 15 51 AM


src/i18n/en/feedback.json
```diff
@@ -12,6 +12,12 @@
     "intakeForm": "By helping us understand the importance of communication to you and your family,  you help us keep the service free for all families impacted by incarceration!",
     "no_show": "You were all set up on your side, but it looks like {{names}} never connected. If you like, you can always check in with them to see what happened."
   },
+  "callEndReason": {
+   "disconnect":"You ended the voice call",
+   "peerDisconnect":"{{visitorFirstName}} ended the voice call",
+   "terminated":"${facilityPublicId} terminated your call. That’s all we know. We apologize for the inconvenience.",
```
cainlevy, line 18: what's the difference between `{{...}}` and `${...}`?

cainlevy (approved):

src/i18n/en/feedback.json
wittejm: Ooh, copy/past fail, good catch. (This copy is untested because DOC-terminate is unsupported atm)

PR #230, opened 2022-04-01
await createallreport

gabrielsaruhashi:
I believe this will address a race condition with the logout call
https://sentry.io/organizations/ameelio/issues/3148331099/?project=6021907&referrer=slack


cainlevy (approved):
i can't figure out how this is related to the sentry ticket, or how it will change pathways behavior. the CallFeedback component is already `await`-ing the promise returned from `rateCall`.

no harm, though.

PR #235, opened 2022-04-05
Refactor/login user

crcommons:


src/graphql/queries/GetMe.graphql
```diff
@@ -0,0 +1,19 @@
+query GetMe {
+  me {
+    ... on Inmate {
+      id
+      malanId
```
cainlevy, line 5: i wonder if we can drop this field as well? i'm hoping we can transition `malanId` into a private implementation detail.

src/App.tsx
```diff
@@ -33,13 +34,14 @@ const App: React.FC<Props> = ({ history }) => {
   const hasSideBar = useAppSelector((state) => state.ui.fullScreen);
   const session = useAppSelector((state) => state.session);
   const dispatch = useAppDispatch();
+  const [token, setToken] = useLocalStorage("token", "");
```
cainlevy, line 37: we've historically only kept the api token in ephemeral page memory. i suspect the decision was originally made because pathways is accessed on shared computers in an environment where we want to discourage account sharing as much as is reasonable.

src/App.tsx
```diff
@@ -50,21 +52,29 @@ const App: React.FC<Props> = ({ history }) => {
...
+            />
             {ROUTES.map((route) => (
               <ProtectedRoute
                 exact
                 {...defaultProtectedRouteProps}
                 path={route.path}
                 component={route.component}
                 key={route.label}
+                token={token}
+                setUser={setUser}
```
cainlevy, line 74: right now it looks like ProtectedRoute needs `setUser` to load the current user's information. what if that responsibility lived here in App.tsx? i have a sense that it may be nice to co-locate the token & current user management, so we don't spread awareness of the difference between having a token and having a session.

src/api/apollo.tsx
```diff
@@ -1,16 +1,18 @@
...
-import { Store } from "../redux";
+// import { Store } from "../redux";
 import { setContext } from "@apollo/client/link/context";
 
 const httpLink = createHttpLink({
   uri: (process.env.REACT_APP_API_URL || "").replace(/v1.*/, "graphql"),
 });
 
 const authLink = setContext((_, { headers }) => {
+  const token = localStorage.getItem("token");
```
cainlevy, line 10: ohhh this must be why you decided on `localStorage`. hmm. i see the challenge but i do think we need a solution that won't persist through page reloads.

src/pages/Login/index.tsx
```diff
@@ -67,19 +76,42 @@ const LoginContainer: React.FC<RouteComponentProps<TParams>> = ({ match }) => {
...
-  if (session.isLoggedIn) {
-    return <Redirect to="/" />;
+  const [loginUser, { error: loginError, loading: loginLoading }] =
+    useLazyQuery(LoginUserDocument, {
+      onError: (e) => {
+        reportApolloError(e, "Login/index/loginUser");
+        Sentry.captureException(e);
+        showToast(
+          "login",
+          "Unable to login at this time.", // how to send login errors vs. connection issues
```
cainlevy, line 87: yeah it seems especially important to identify credentials issues

src/App.tsx
crcommons: There is an issue that because app will try to run the user query even though you're not logged in. I attempted to move it to app and only run the query if we have the token, but there were some issues with the infinite re-render.

cainlevy (changes requested):
i think we're getting close!

please use your discretion on some of these. i'm most interested in maintaining behavior on page reloads with a tweak to `useSessionStorage`.

in a future (unplanned) iteration we could also consider:

- React context
- revisiting the routes list

src/App.tsx
```diff
@@ -32,39 +39,94 @@ const App: React.FC<Props> = ({ history }) => {
...
   };
-
-  const showSideBar = session.isLoggedIn && !hasSideBar;
+  const showSideBar = token && !hasSideBar;
 
   useEffect(() => {
     i18n.changeLanguage(session.language);
   }, [session.language, i18n]);
 
+  // This needs to be defined within a react component
```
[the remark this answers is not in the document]
crcommons: Yes and because you can't provide a jsx value outside of react

src/pages/Login/index.tsx
```diff
@@ -67,19 +74,42 @@ const LoginContainer: React.FC<RouteComponentProps<TParams>> = ({ match }) => {
...
+      },
+    });
+
+  let history = useHistory();
+
+  if (error || loginError) throw error;
+
+  if (props.token) {
+    history.push("/");
   }
```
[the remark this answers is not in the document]
crcommons: That's one reason, but it existed before because if you navigate to /login, we need to push you back to the home page. I just needed to make the change because 1) we're not using session.isLoggedIn anymore and 2) I read the it's preferable to use history.push over the redirect component.

src/pages/CallFeedback.tsx
```diff
@@ -13,20 +13,22 @@ import Loader from "src/components/Loader";
 import { CreateCallReportDocument } from "src/graphql/mutations/CreateCallReport.generated";
 import { GetMeetingDocument } from "src/graphql/queries/GetMeeting.generated";
 import { useAppDispatch } from "src/redux";
-import { logout } from "src/redux/modules/session";
 import { enterFullScreen, exitFullScreen } from "src/redux/modules/UI";
 import { hasMeetingExpired } from "src/utils";
 
-type TParams = { id: string; exitType: CallExitType };
+// Am I going to get in trouble for making these option?
+type TParams = { id?: string; exitType?: CallExitType };
```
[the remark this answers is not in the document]
crcommons: Yes :(

Maybe we can look at it later today

src/hooks/useAuthInfo.ts
```diff
@@ -0,0 +1,36 @@
...
+      query,
+      data: {
+        authInfo: {
+          __typename: "AuthInfo",
+          ...value,
+        },
+      },
+    });
+    setValue(value);
+  };
```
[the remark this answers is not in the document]
crcommons: that was a different problem with usestate. This hook has been working like a dream :)

cainlevy (approved):

PR #236, opened 2022-04-05
Refactor/user data cache

crcommons:


src/components/Profile/ProfileInfo.tsx
```diff
@@ -4,49 +4,56 @@ import { format } from "date-fns";
...
 interface Props {
-  user: User;
   onEdit: () => void;
 }
 
-const ProfileInfo: React.FC<Props> = ({ user, onEdit }) => {
+const ProfileInfo: React.FC<Props> = ({ onEdit }) => {
+  const { user, userLoading } = useUser();
   const { t } = useTranslation("profile");
+  const fullName = getFullName(user);
```
cainlevy, line 19: is this safe to call while `userLoading`?

come to think of it, should this page ever be opened while the user is loading? or should we resolve that somewhere else and add some kind of guarantee (e.g. error condition) to `useUser()`?

src/hooks/useUser.ts
```diff
@@ -0,0 +1,23 @@
+import { useQuery } from "@apollo/client";
+import { GetMeDocument } from "src/graphql/queries/GetMe.generated";
+import { Inmate } from "src/types/User";
+
+const useUser = () => {
+  const {
+    data: { me } = {},
+    loading: userLoading,
+    error: userError,
+  } = useQuery<any>(GetMeDocument, {
```
cainlevy, line 10: on the one hand, i see the appeal of loading `me` precisely where needed. on the other hand, it will mean that "is the user loaded yet" becomes a problem _everywhere_. i anticipate that could result on a lot of spinners, even in places where it's not possible to access the component without resolving the current user.

PR #237, opened 2022-04-08
render participant comments for voice calls

wittejm:
Screen Shot 2022-04-08 at 2 53 48 PM


src/components/Inbox/VoiceCallCard.tsx
```diff
@@ -31,8 +31,14 @@ const VoiceCallCard = ({ meeting, className }: Props) => {
           <span className="text-gray-500 text-sm">
             {`${
               meeting.status === MeetingStatus.NoShow ? "Missed " : ""
-            } ${getVisitLabel(meeting.meetingType)} •
-            ${millisecondsToHoursOrMinutesLabel(meeting.call?.duration)}`}
+            } ${getVisitLabel(meeting.meetingType)}
+            ${
+              meeting.status === MeetingStatus.NoShow ||
```
wittejm, line 36: an unrelated fix: don't show the call duration if there shouldn't be one.

cainlevy (approved):

PR #239, opened 2022-04-12
test disconnect using local build/pack/install of ccc

wittejm:
Calling `disconnect` seems to stop the mobile->pathways audio but doesn't stop the pathways-> mobile audio


cainlevy:
> but doesn't stop the pathways-> mobile audio

oh what. i wonder how that works. 🤔

PR #240, opened 2022-04-12
disconnect call on end call mutation

wittejm:
This works locally! Woohoo!


src/components/Inbox/VoiceCall/VoiceCallInterface.tsx
```diff
@@ -50,11 +50,13 @@ const VoiceCallInterface = ({ vc, authInfo, onFinish }: Props) => {
   const [endTime, setEndTime] = useState<Date | null>(null);
   const [callEndReason, setCallEndReason] =
     useState<CALL_END_REASON>("disconnect");
+  const disconnectRef = useRef<() => Promise<void>>();
```
cainlevy, line 53: i was going to suggest that the hooks could also be re-ordered but i remember you observing that they're getting rather entangled 😬

cainlevy (approved):

src/components/Inbox/VoiceCall/VoiceCallInterface.tsx
wittejm: Yep, in this case the dependency loop is disconnect() >> endCall mutation >> handlePeerDisconnect >> useConnectCall >> disconnect()

PR #243, opened 2022-04-13
Show seconds in all call durations

wittejm:
A few in-line examples: 

Screen Shot 2022-04-13 at 2 53 03 PM


src/utils/Time.spec.ts
```diff
@@ -0,0 +1,62 @@
+import { millisecondsToHoursMinutesSecondsLabel } from "./Time";
+
+describe("millisecondsToHoursMinutesSecondsLabel", () => {
+  it("formats a bunch of different things correctly", async () => {
```
cainlevy, line 4: 😆

cainlevy (approved):
hard to argue with test coverage like that

PR #245, opened 2022-04-14
seconds-specific call durations in pre- and post-call modals

wittejm:


cainlevy (approved):
the number of derived values in StartVoiceCallModal is getting to be a little intimidating!

wittejm:
Agreed!

PR #248, opened 2022-04-19
Show "you ran out of time" closeout text

wittejm:
Tweaking the time variables and variable names a little bit here too

Screen Shot 2022-04-19 at 2 59 43 PM


src/components/Inbox/VoiceCall/VoiceCallInterface.tsx
```diff
@@ -128,12 +128,25 @@ const VoiceCallInterface = ({ vc, onFinish }: Props) => {
         setRenderState("ended");
         break;
       case "ended":
-        setRenderState("missed");
+        if (endTime && endTime.getTime() - Date.now() < 0) {
+          setCallEndReason("timeUp");
```
wittejm, line 132: This is the only pertinent logic for the PR behavior change. Pathways has to decide for itself whether the call ran to its maximum allowed end time, so we check it here.

cainlevy (approved):

PR #253, opened 2022-04-22
sc-3370 added toasts/copy for connection state changes

jackofsometrade:


jackofsometrade:
Oops looks like the test action will fail until we merge and release a new connect-call-client...

src/components/Call/Call.tsx
```diff
@@ -120,6 +132,108 @@ const CallScreen: React.FC<Props> = React.memo(
...
       onPeerDisconnected: handlePeerDisconnected,
       onNewMessage: handleNewMessage,
+      onConnectionState: throttle((state: ConnectionState) => {
+        switch (state.quality) {
+          case "poor":
+            if (currentConnectionQuality !== "poor" && !videoDisabled) {
+              const poorToast = () => {
+                showToast(
+                  "poorConnectionState",
+                  t("call:connectionQuality.poor"),
```
wittejm, line 142: What do you think of using this new approach of i18n I'm trying? Content text as translation keys so we can still see content language in-line: 
https://github.com/Ameelio/pathways-client/blob/187a760f3293b566e7931bc41364700e8e8fd667/src/components/Inbox/VoiceCall/StartVoiceCallModal.tsx#L66

src/components/Call/Call.tsx
```diff
@@ -62,6 +65,15 @@ const CallScreen: React.FC<Props> = React.memo(
...
     const user = useUser();
+    const [videoDisabled, setVideoDisabled] = useState(false);
+    const [currentConnectionQuality, setConnectionQuality] =
+      useState<ConnectionState["quality"]>("excellent");
+    const [currentPeerConnectionQuality, setPeerConnectionQuality] =
+      useState<ConnectionState["quality"]>("excellent");
+    const [deferredPoorQualityToast, setDeferredPoorQualityToast] =
+      useState<() => void>();
+    // const localVideoRef = useRef<VideoTrack>();
+    // const toggleVideoRef = useRef<() => void>();
```
wittejm, line 76: What are these?

wittejm (dismissed):

src/components/Call/Call.tsx
```diff
@@ -120,6 +132,108 @@ const CallScreen: React.FC<Props> = React.memo(
...
+        switch (peerState.quality) {
+          case "poor":
+            if (currentPeerConnectionQuality !== "poor") {
+              const peerFirstName = call.visitors[0]?.firstName;
+              showToast(
+                "poorConnectionState",
+                t("call:connectionQuality.peerPoor").replace(
+                  /PEER_FIRST_NAME/g,
+                  peerFirstName
+                ),
```
cainlevy, line 204: https://www.i18next.com/translation-function/interpolation

src/components/Call/Call.tsx
jackofsometrade: Leftover pasta from connect-mobile, deleting.

src/components/Call/Call.tsx
```diff
@@ -218,6 +231,114 @@ const CallScreen: React.FC<Props> = React.memo(
       [sendMessage, setErroredMessages, authInfo]
     );
 
+    useEffect(() => {
+      switch (connectionState.quality) {
+        case "poor":
+          if (!videoDisabled) {
```
cainlevy, line 237: i find it easier to read code with an early return because i don't have to scan for the matching brace and mind the indentation, e.g.:

```
if (!videoDisabled) break;
```

src/components/Call/Call.tsx
```diff
@@ -62,6 +69,11 @@ const CallScreen: React.FC<Props> = React.memo(
     const [playLeaveCall] = useSound(LeaveCallSound);
     const [playMessageReceived] = useSound(MessageReceivedSound);
     const user = useUser();
+    const [videoDisabled, setVideoDisabled] = useState(false);
+    const connectionQualityRef = useRef<Quality>();
+    const peerConnectionQualityRef = useRef<Quality>();
```
cainlevy, line 74: i saw a hook in cmob that might help relocate some of this complexity. it was a version of `useEffect` that would be triggered by certain value transitions.

the exact implementation may not be perfect but the concept might be really helpful:

https://github.com/Ameelio/connect-mobile/blob/main/src/utils/hooks/useChange.ts

imagine something like:

```
useTransition({to: "bad"}, () => {
});

useTransition({from: ["bad", "poor"], to: ["good", "excellent", "average"]}, () => {
});
```

cainlevy (changes requested):
really close, i promise!

i'm mostly wary of how much transition effect awareness is built into the useEffect that watches the current user's connection state. the issue, i think, is how the callback is responsible for properly maintaining a `videoDisabled` value. let's talk and see if we can refine the pattern!

everything else is easily deferred polish or code golf.

src/components/Call/VideoOverlay.tsx
```diff
@@ -15,6 +15,7 @@ interface Props {
   audioOn: boolean;
   toggleAudio: () => void;
   videoOn: boolean;
+  videoDisabled: boolean;
```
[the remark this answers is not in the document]
jackofsometrade: yeah we can change it to something more obvious

src/components/Call/Call.tsx
```diff
@@ -218,6 +231,114 @@ const CallScreen: React.FC<Props> = React.memo(
       [sendMessage, setErroredMessages, authInfo]
     );
 
+    useEffect(() => {
```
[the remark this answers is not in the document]
jackofsometrade: Sure thing, that helps me too.

src/components/Call/Call.tsx
```diff
@@ -218,6 +222,86 @@ const CallScreen: React.FC<Props> = React.memo(
...
+    useChange(connectionState.videoDisabled, (_previous, videoDisabled) => {
+      if (videoDisabled) {
+        showToast(
+          "connectionState",
+          !localVideo?.paused
+            ? t("call:connectionQuality.bad")
+            : t("call:connectionQuality.badWhenVideoPaused"),
+          "error"
+        );
+        if (localVideo && !localVideo.paused) localVideo.paused = true;
```
cainlevy, line 272: 🤔 why mutate `localVideo`? won't ccc's RoomClient do that?

cainlevy (approved):
wow `useChange` had a bigger impact than i expected!

(i'm resisting the temptation to play golf with the imported code 🤐 )

src/components/Call/Call.tsx
jackofsometrade: It only seems to do that when setLocalVideo is called (in ccc) which only sets paused = true when the user initiates a `toggleVideo` command.  This is represents some of the state stuff that I think could use a rethink.

cainlevy: oh. sounds like RoomClient may need some additional logic for the new situation?

mutating react state is an anti-pattern because it does not trigger re-renders. it might also not persist between renders.

jackofsometrade: Yeah I can see that now, I hadn't realized this before. I'll make some adjustments, thanks!

jackofsometrade:
@cainlevy Ping, hopefully last change needed :pray:

cainlevy (approved):

PR #254, opened 2022-04-25
Jw/sc 3528/grammar corrections in minutes rules in call

wittejm:
Escaping periods initially didn't render in i18n, and then they did for no discernible reason, and then it stopped working again for no discernible reason, so my new approach is to represent periods in translation keys with "__". It still kind of sucks, but I still like this *so much more* than using arbitrary keys for translations.

Screen Shot 2022-04-25 at 9 42 13 AM


cainlevy (approved):
> my new approach is to represent periods in translation keys with "__". It still kind of sucks, but I still like this _so much more_ than using arbitrary keys for translations.

that says a lot!

you're passionate about finding a good workflow here so i want to be encouraging. however, this adds to my concern by demonstrating that we're working _against_ the library (and, thereby, the convention of everyone using it). i don't think this should be our final iteration on how we manage translations.

wittejm:
> i don't think this should be our final iteration on how we manage translations.

Fair! The first time I worked with in-page translations was with [ttag](https://ttag.js.org/docs/quickstart.html) which is intended to work this way. I can't say that I worked with it (or i18n generally) long enough to feel confident that a refactor would be purely beneficial, also ofc a refactor would be time intensive.

PR #255, opened 2022-04-26
reset countdown timer for minutes available

wittejm:
Depends on https://github.com/Ameelio/connect-api-nest/pull/872
Screen Shot 2022-04-26 at 3 51 47 PM
Screen Shot 2022-04-26 at 3 54 47 PM


cainlevy (approved):
it looks like this feature assumes synchronized clocks, so that the client can calculate the distance to a date returned by the server. that assumption will fail us a surprising amount of the time. in this case i think it would cause our product to be perceived as inconsistent.

a countdown like this (especially a live one) invites the user to wait in anticipation and act when the time is finished. however i think this can bring disappointment and frustration in two ways:

1. the interface (as built) will not update dynamically when the timer finishes its countdown. that is, they will not see the available minutes change while they watch.
2. in the case of desynchronized clocks where the client's clock is ahead, the server may not even agree that the countdown is finished.

both of these are fixable with additional effort, but it's worth checking on the cost/benefit here.

cainlevy:
@cshdllr it's really hard to estimate how often this will happen. my guess is that the incidence rate would vary by device type. phones, for example, seem to have well-synchronized clocks whereas windows machines have a reputation (in my mind at least) of clock drift. i couldn't guess where chromebooks fall or what the effect of IT management might be.

one way to fix it is by returning a duration instead of a date. the down side is that the duration is guaranteed to be a little bit stale by the time our client gets it. the up side is that it would be a consistently small amount wrong (< 1s) instead of the rarely large amount wrong (> 1m) in the case of bad clock drift.

cshdllr:
Based on our conversation today, I've updated the designs to be a specific time instead of the countdown. 

Designs can still be found here: https://www.figma.com/file/IXcl41KdiXJRP4QRx7TKHU/?node-id=264%3A29422

image

wittejm:
Just awaiting a nitpick question here: https://ameelioorg.slack.com/archives/C01TGJH69V4/p1651086095396779?thread_ts=1651084231.667849&cid=C01TGJH69V4

PR #257, opened 2022-04-29
Timer flips countup to countdown at 5 minutes left

wittejm:
In both voice and video!

https://user-images.githubusercontent.com/2104990/166058242-658d0752-7db3-414d-9aa8-2aa864aa4529.mov

https://user-images.githubusercontent.com/2104990/166058142-c46fbf0e-63f5-447a-a54e-ae1e28eddffb.mov


cainlevy (approved):
took me a bit to guess why we have separate implementations for the countdown and the countup. without getting into those peculiarities, i wonder if the `<Timer />` component should be entirely responsible for both? right now it owns the countdown, based on `endTime`, but must be passed the countup via `timerValue`.

cainlevy:
By the way I love the reuse here! Seems like a nice improvement to video calls.

wittejm:
Yeah it's a little patched together. I pondered enclosing timerValue inside timer at some point, but a solution there was not obvious. Also Countdown provides us with `tick` events that trigger the up/down flips, so I had to leave that intact and work within its limits (it only takes the endTime and it wouldn't know from where to begin a count up).

This could certainly be improved! But I'm sticking a fork in it.

PR #258, opened 2022-05-02
customize antd color

wittejm:
First step to fixing colors is to update Antd's blue primary color. Links and buttons look like this now:

Screen Shot 2022-05-02 at 8 33 57 AM

This npm dependency didn't install without using `--legacy-peer-deps` and it updated package-lock.json ... bumping a few dependency versions and also replacing `"devOptional": true` with `"dev": true`. What are possible consequences or concerns with this?


cainlevy:
What dependencies caused problems?

The risk is that we ignore an API incompatibility between our library and its expected dependencies. Possible resolutions:

- find an older version of the library that's compatible
- upgrade our blocking dependencies
- test thoroughly and leave this for the next person to figure out

wittejm:
Cool, I went and learned a bit about what peer dependencies are. The only version conflict is craco-less's peer dependency of react-scripts 5.0.1; ours is 4.0.0. Out of caution I bumped the craco-less version down from 2.0.0 to 1.20.0. Unfortunately their repo has no changelog 😭 . But it still works for my minimal test of replacing 1 color (antd's primary). So I'm willing to proceed with this.

wittejm:
Calling this a draft branch because I'm putting in our custom color palette here too. This early PR was just for thoughts on the peer dependencies thing 👍

cainlevy (approved):

PR #260, opened 2022-05-03
Fix: Only show first time user modal once

wittejm:
Commits 1 and 3 are just to show my work of messing around before I got this working -- my first attempt at a cache modify wasn't working because `currentInmateId` is `undefined` at the time of the executed mutation's `onCompleted` function. That doesn't make sense? Does it use the value at the function's definition time rather than its execution time? I feel like that's a really naive JS scoping question. 

Fortunately, the inmate id is conveniently available in the mutation response, so that works fine.


src/pages/InboxPage.tsx
```diff
@@ -125,6 +125,16 @@ const InboxPage: React.FC<Props> = () => {
             },
           },
         });
+        cache.modify({
+          id: cache.identify({
+            __typename: "Inmate",
+            id: message.sender.id,
```
cainlevy, line 131: > Does it use the value at the function's definition time rather than its execution time?

Yep! That's the closure. `useRef` is one way out.

cainlevy (approved):

PR #261, opened 2022-05-04
Jw/sc 3559/update minutes at end call

wittejm:
This updates the call times the next time the user opens the start-call modal, without needing a page refresh.

The fact that the fix is this simple is really cool.


cainlevy (approved):

cainlevy:
@wittejm is there still an issue with the available minutes being out of date when an inmate opens the modal? because the data is loaded before.

wittejm:
This is for the case that the user tries to make a second call: the modal content doesn't currently reflect the time reduced by completing the first call.

cainlevy:
Yes, I understand. My memory of the implementation though is that the modal shows how many minutes were available when the page was loaded, even if the user waits to open the modal. I was asking if we're still tracking that one. It may may this pull request's solution redundant.

PR #263, opened 2022-05-05
Tywarren/sc 3362/bad post call ux

jackofsometrade:


src/components/CallFeedback/CallFeedback.tsx
```diff
@@ -72,11 +75,38 @@ const NoShowFeedback = ({
...
   const { t } = useTranslation(["feedback", "common"]);
 
   const navigate = useAppNavigate();
 
+  const { droppedStreams } = useDroppedStreams(call.id);
+  const { data: facilityNameData } = useQuery(GetMyFacilityNameDocument);
+
+  const badCallDescription = useMemo(() => {
+    switch (true) {
+      case droppedStreams.local && !droppedStreams.peer:
```
crcommons, line 93: Is it possible for local and peer to both evaluate to true?

src/components/CallFeedback/CallFeedback.tsx
```diff
@@ -72,11 +75,38 @@ const NoShowFeedback = ({
...
+
 const CallFeedback: React.FC<Props> = ({ call, rateCall, type, logout }) => {
   const { t } = useTranslation(["feedback", "common"]);
 
   const navigate = useAppNavigate();
 
+  const { droppedStreams } = useDroppedStreams(call.id);
+  const { data: facilityNameData } = useQuery(GetMyFacilityNameDocument);
+
+  const badCallDescription = useMemo(() => {
```
crcommons, line 91: Nice use of memoize here 😎

crcommons (approved):
Nice job!

src/components/Call/Call.tsx
```diff
@@ -64,6 +65,7 @@ const CallScreen: React.FC<Props> = React.memo(
     const user = useUser();
     const [deferredPoorQualityToast, setDeferredPoorQualityToast] =
       useState<() => void>();
+    const { droppedStreams, setDroppedStreams } = useDroppedStreams(call.id);
```
cainlevy, line 68: this hook's api could be simpler if it exported `addDroppedStream` instead of `setDroppedStreams`. that would let it encapsulate the correct way of adding, and we wouldn't need to reference `droppedStreams` here.

src/components/CallFeedback/CallFeedback.tsx
```diff
@@ -72,11 +75,38 @@ const NoShowFeedback = ({
...
+      case !droppedStreams.local && droppedStreams.peer:
+        return t("feedback:badCall.peer", {
+          peerFirstName: getFirstNames(call.visitors),
+        });
+      case droppedStreams.local && droppedStreams.peer:
+        return t("feedback:badCall.localAndPeer", {
+          facilityName:
+            facilityNameData?.currentInmate?.facility.name || "facility",
+        });
+    }
```
cainlevy, line 107: would it make sense for `useDroppedStreams` to compute and return a primitive like `who: 'local' | 'peer' | 'localAndPeer'`? it doesn't look like we actually need any other information on `droppedStreams`.

cainlevy (commented):
i wonder if this feature could have been implemented with a param to the feedback page like `droppedStreams: 'me' | 'them' | 'us'`.

jackofsometrade:
Ok, this is quite different, but in exchange for a little bit of esoteric code, we get really simple usage, and string literals where it counts.

src/components/Call/Call.tsx
```diff
@@ -42,12 +42,28 @@ type Call = GetCallQuery["meeting"];
...
+export type StreamDropType = typeof streamDropTypes[number];
+const DroppedStreams: Record<StreamDropType, number> = {
+  [streamDropTypes[0]]: 0x00,
+  [streamDropTypes[1]]: 0x01,
+  [streamDropTypes[2]]: 0x02,
+  [streamDropTypes[3]]: 0x03,
+};
+const getDropType = (value: number): StreamDropType => {
+  return (
+    (Object.keys(DroppedStreams) as StreamDropType[]).find(
```
cainlevy, line 61: nice benefit of the `as const` enum pattern:

```suggestion
    streamDropTypes.find(
```

cainlevy (approved):
i've always wanted to ship a bit field to production!

PR #264, opened 2022-05-05
remove page double scrollbar

wittejm:
Sets the correct component to the max screen size rather than a child component which was making the parent too large
Screen Shot 2022-05-05 at 9 17 06 AM
.


cainlevy (approved):
ahhh relief!

PR #281, opened 2022-05-16
Jw/sc 3727/start call modal fix time format

wittejm:


cainlevy (approved):
let's consider adding or creating a small duration lib to the providers client! imagine passing around an object that knows its unit and can manage simple conversions:

```
const d = new Duration({ms: 123456})
console.log(d.seconds) // 123.456
console.log(d.minutes) // 2.0576
```

wittejm:
That sounds great! This time formatting stuff has gotten comical.

PR #285, opened 2022-05-16
bugfix: make modal not closable while call loads

wittejm:
Killing the call while it's initializing is not something I want to get into. Without this fix, we could stay in the "calling" state with the call modal closed and the call button disabled. The call then completes and brings up the in-call modal. Let's disallow this.


cainlevy (approved):
i wonder if this deserves an inline comment to warn the next person about the complexity we're avoiding

PR #287, opened 2022-05-21
when mobile declines the voice call, pathways handles it like a missed call

gabrielsaruhashi:
Right now mobile keeps dialing even after Pathways declines the voice call


cainlevy (approved):

crcommons (approved):

wittejm:
This server event could also include the instant reply, correct? Could those two piece of logic (updating to render the instant reply) be merged here?

PR #294, opened 2022-08-01
POC Cisco integration

dabbler0:


package.json
```diff
@@ -7,19 +7,47 @@
...
+    "@webex/storage-adapter-local-storage": "1.157.0",
+    "@webex/webex-core": "1.157.0",
     "antd": "^4.16.13",
     "apollo-link-sentry": "^3.1.1",
     "assert-never": "^1.2.1",
     "autoprefixer": "^9.8.6",
     "camelcase-keys": "^6.2.2",
     "connect-call-client": "^0.9.0",
     "connected-react-router": "^6.8.0",
+    "core-js": "^3.24.1",
```
cainlevy, line 50: is core-js a primary dependency now or was there a versioning issue?

package.json
```diff
@@ -7,19 +7,47 @@
     "@ant-design/icons": "^4.7.0",
     "@apollo/client": "^3.5.5",
     "@craco/craco": "^6.1.1",
+    "@graphql-typed-document-node/core": "^3.1.1",
```
cainlevy, line 10: what was the motivation to make this a primary dependency?

package.json
```diff
@@ -113,6 +142,6 @@
     "lint-staged": "^11.2.6",
     "prettier": "^2.4.1",
     "prettier-plugin-organize-imports": "^2.3.4",
-    "typescript": "^4.4.4"
+    "typescript": "^4.1.5"
```
cainlevy, line 145: why downgrade?

src/components/Call/CiscoCall.tsx
```diff
@@ -0,0 +1,500 @@
...
+  const [decryptedConfig, setDecryptedConfig] = useState<WebexConfig>();
+  const [webex, setWebex] = useState<WebexInstance>();
+  const [meeting, setMeeting] = useState<WebexMeetingInstance>();
+  const [remoteVideo, setRemoteVideo] = useState<WrappedMediaStream>();
+  const [remoteAudio, setRemoteAudio] = useState<WrappedMediaStream>();
+  const [localVideo, setLocalVideo] = useState<WrappedMediaStream>();
+
+  useEffect(() => {
+    if (!madeDecryptionRequest) {
+      setMadeDecryptionRequest(true);
```
cainlevy, line 130: i get the impression that this and the following `useEffect` hooks implement a sequence by updating state variables in a way that causes the next step to happen just once.

if that's the case, i think it might be easier to follow with a single handshake or connection function. we might get to drop a lot of intermediate state and spend less time thinking about `useEffect` dependency chains.

src/components/Call/CiscoCall.tsx
```diff
@@ -0,0 +1,500 @@
...
+  leave: () => Promise<void>;
+  addMedia: (o: { localStream?: WrappedMediaStream, localShare?: WrappedMediaStream, mediaSettings?: MediaOptions}) => void;
+  getMediaStreams: (mediaSettings: MediaOptions) => WrappedMediaStream[];
+  muteVideo: () => Promise<void>;
+  unmuteVideo: () => Promise<void>;
+  muteAudio: () => Promise<void>;
+  unmuteAudio: () => Promise<void>;
+}
+
+const useWebexCall = ({
```
cainlevy, line 104: probably worth extracting to `src/components/Call/useWebexCall.ts`. it would allow the pieces that remain to be more clearly related to `src/components/Call/Call.tsx` (and maybe eventually reconciled).

cainlevy (commented):
😂 image

there's probably a lot of opportunity to reconcile the duplication between CiscoCall.tsx and Call.tsx but i'm personally more invested in getting that right with the Providers codebase.

i'm most interested in understanding the necessity of some dependency changes in package.json. other aspects of readability could happen in a follow up (or in Providers).

package.json
dabbler0: There was a versioning issue. Something got out of sync when I added all the new dependencies so I removed `package-lock.json` and reinstalled directly from `package.json`. When I did that, a bunch of things broke. I then just kept adding individual explicit dependencies until things started working again.

package.json
dabbler0: Same as above. For some reason I don't understand, when I removed `package-lock.json` and then npm installed, this didn't get included. So I included it explicitly.

package.json
cainlevy: oh that explains why package-lock.json has so many changes: re-creating it effectively updates the entire project's list of dependencies within the constraints of the package.json specifications. it means that deploying requires confidence that the entire application works, not just the cisco changes in this branch.

since pathways is a fairly small application, that probably just means verifying that CVH video works and attempting to undo these secondary changes to the dependency list.

in another application, it would probably be worth starting fresh with the dependencies and trying to figure out a more targeted resolution.

cainlevy (approved):

PR #301, opened 2022-10-04
Sc 4694

enrjp:
Removed the InPersonNoContact value from a few enums. Application compiles successfully after doing so. Also removed the InPersonContact type which is no longer needed.


cainlevy (approved):
thinking about backwards compatibility here. i'm pretty sure this will work because the value is unused, so we should never be surprised by an unhandled value.

if we continue this pattern in other repositories we'll need to make sure everyone is aware that they should _not_ accept all changes to `src/api/graphql.ts`. doing so would introduce type errors in all of the places that expect an exhaustive list of meeting types.

alternatively, you could look through the remaining clients for any situations that would result in an error if this enum value was no longer supported by the server. if we're confident we can remove the graphql enum without breaking any clients, we can unpublish it from the api first and then take the steps shown here as cleanup.

what do you think?

enrjp:
>

enrjp:
@cainlevy so my intention was to go through all the clients and remove any references to the enum value. But probably best to leave the schema untouched for now, deploy to staging, and perform appropriate regression testing. Once we've confirmed all the clients are error-free and (go to Prod?) we could look at removing from the API and re-generate/re-importing into the clients.

Sorry if my order-of-operations might be slightly off, still getting up-to-speed on the ideal workflow.

cainlevy:
Let's go for it and see what we learn. I think this is a safe step.

> But probably best to leave the schema untouched for now, deploy to staging, and perform appropriate regression testing. Once we've confirmed all the clients are error-free and (go to Prod?)

When you merge this pull request, it will become the current version in both staging and production. It will not pause for QA on staging or synchronize deploys alongside any other repository. That's why we aim to have fully functional local environments and make every change backwards compatible.

If you're wondering why the system works this way, let's talk it through in person tomorrow!

enrjp:
Sounds good, thanks @cainlevy !

PR #302, opened 2022-10-05
remove references to MeetingType.InPersonNoContact

enrjp:
Updated the branch/PR to only remove the references/implementations of the enum value, but not delete the value from the graphql schema.


src/components/Meetings/Meetings.tsx
```diff
@@ -52,13 +52,6 @@ const Meetings: React.FC<Props> = ({ meetings, loading }) => {
           meetingType={MeetingType.InPersonContact}
         />
       </Menu.Item>
-      <Menu.Item>
```
enrjp, line 55: Removing the menu item (should no longer be visible)

src/types/Meeting.ts
```diff
@@ -46,14 +45,7 @@ export type VoiceCall = Omit<Meeting, "meetingType" | "call"> & {
...
   meetingType: MeetingType.InPersonContact;
   call: null;
 };
-
-type InPersonNoContactVisit = Omit<Meeting, "meetingType" | "call"> & {
-  meetingType: MeetingType.InPersonNoContact;
-  call: null;
-};
-
-export type InPersonVisit = InPersonContactVisit | InPersonNoContactVisit;
```
enrjp, line 59: No longer need this aggregate type - can just use InPersonContactVisit (so I exported it above)

src/types/Meeting.ts
```diff
@@ -6,11 +6,10 @@ import {
...
 import { VisitorFragmentFragment } from "src/graphql/fragments/VisitorFragment.generated";
 
-export function meetingIsVisit(meeting: Meeting): meeting is InPersonVisit {
-  return (
-    meeting.meetingType === MeetingType.InPersonContact ||
-    meeting.meetingType === MeetingType.InPersonNoContact
-  );
+export function meetingIsVisit(
+  meeting: Meeting
+): meeting is InPersonContactVisit {
```
enrjp, line 11: InPersonVisit becomes just InPersonContactVisit

src/components/Meetings/MeetingsTable.tsx
```diff
@@ -36,8 +36,6 @@ const MeetingsTable: React.FC<Props> = ({
         return <VideoCameraGray />;
       case MeetingType.InPersonContact:
         return <PeopleGray />;
-      case MeetingType.InPersonNoContact:
```
enrjp, line 39: Shouldn't need to handle this case anymore? I suppose if there were meeting records in the DB that were of type NoContact it would display the default case VideoCameraGray component. Can add it back if desired.

src/utils/utils.ts
```diff
@@ -26,8 +26,6 @@ export const getVisitLabel = (
       return "Video Call";
     case MeetingType.InPersonContact:
       return ignoreContact ? "In-Person Visit" : "In Person, With Contact";
-    case MeetingType.InPersonNoContact:
```
enrjp, line 29: Same as above - shouldn't need to handle this case anymore I think, I suppose if there were meeting records in the DB that were of type NoContact it would display the default case VideoCameraGray component. Can add it back if desired.

src/components/Meetings/MeetingsTable.tsx
cainlevy: i've confirmed that we don't have any of these values in the db

cainlevy (approved):
the remaining mentions of `InPersonNoContact` look harmless to me

enrjp:
Closing pull request, will start fresh once the updated API has been released

